PR #13, opened 2023-05-05
Updates to support the new HiveMQ plugin.

amrc-benmorrow:
These changes will only function with the new plugin (my open PR).

I think perhaps the chart should pull in fixed versions of the components, rather than `:latest`? We maybe also need a policy about whether a change to the required configuration is a semver major, minor or patch change.

We need an image that contains `k5start`, and ideally a Kerberised `curl` too. Currently I've pulled `factoryplus-shell` from the internal registry, but I don't think that's the right answer, so this is currently draft.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -109,7 +109,7 @@ mqtt:
     # -- The repository of the MQTT component
     repository: acs-mqtt
     # -- The tag of the MQTT component
-    tag: v1.0.1
+    tag: v1.1.0
     # @ignore
     pullPolicy: IfNotPresent
 
--- a/values.yaml
+++ b/values.yaml
@@ -209,6 +209,13 @@ edo:
     pullPolicy: IfNotPresent
   verbosity: "ALL,!service,!token"
 
+shell:
+  image:
+    registry: ghcr.io/amrc-factoryplus
+    repository: acs-krb-utils
+    tag: v0.0.1
+    pullPolicy: IfNotPresent
+
 minio:
   # -- Whether or not to enable MinIO
   enabled: true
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -130,6 +130,13 @@ warehouse:
       # @ignore
       pullPolicy: IfNotPresent
 
+shell:
+  image:
+    registry: amrc-factoryplus.shef.ac.uk:5000
+    repository: factoryplus-shell
+    tag: 0.0.6
+    pullPolicy: IfNotPresent
```
AlexGodbehere, line 138: @amrc-benmorrow how about this?

```suggestion
    registry: ghcr.io/amrc-factoryplus
    repository: acs-krb-utils
    tag: 1.0.0
    pullPolicy: IfNotPresent
```

templates/_helpers.tpl
```diff
@@ -35,6 +35,14 @@ Create chart name and version as used by the chart label.
...
 {{- end }}
 
+{{/*
+Define the image for a container.
+*/}}
+{{- define "amrc-connectivity-stack.image" -}}
+image: {{ .image.registry }}/{{ .image.repository }}:{{ .image.tag }}
+imagePullPolicy: {{ .image.pullPolicy }}
+{{- end }}
+
```
AlexGodbehere, line 45: What does this do?

AlexGodbehere (commented):
Comments inline. Looks good to me.

templates/_helpers.tpl
amrc-benmorrow: This block is used in `templates/mqtt/mqtt.yaml` to pull in the image references. Factoring it out into a block avoids repeating the logic for the registry/repo/tag/pullPolicy every time.

PR #17, opened 2023-06-27
Enable all components to be enabled or disabled

AlexGodbehere:
This PR enables all existing components to be toggled on/off. It does not include the EDO or Git component.

### DIFF
--- a/templates/auth/principals/operators.yaml
+++ b/templates/auth/principals/operators.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.identity.enabled }}
 # Operator (cluster administration) Kerberos principals.
 
 # This principal has root level access to some services, in particular the authentication database (if enabled).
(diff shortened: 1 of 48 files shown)


templates/auth/principals/operators.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.identity.enabled }}
```
amrc-benmorrow, line 1: I think these principal entries should be enabled based on the individual services that need them, rather than all being enabled with `identity`. Otherwise we can't migrate individual services across.

amrc-benmorrow (changes requested):

AlexGodbehere:
Good spot.

templates/auth/principals/operators.yaml
AlexGodbehere: Agreed, good spot.

amrc-benmorrow (approved):

PR #22, opened 2023-06-30
Add default value for identity.crossRealm to values.yaml

AlexGodbehere:


amrc-benmorrow (changes requested):
`crossRealm` is an array, so both `{}` should be `[]`.

AlexGodbehere:
> `crossRealm` is an array, so both `{}` should be `[]`.

@amrc-benmorrow AFAIK, `{}` means "no defined value" in Helm charts, although I could be wrong.

amrc-benmorrow:
A `values.yaml` is plain YAML, so `{}` is an empty object. Helm treats this as 'empty', along with `null`, `0`, `""` and `[]`, so `default` and `coalesce` will treat an empty object as value-not-there. I think the merging of different `values.yaml` files happens the same way, but this isn't actually documented properly. However, if any non-empty value is supposed to be an array, it seems clearer to me to specify the default as an empty array rather than an empty object.

AlexGodbehere:
> A `values.yaml` is plain YAML, so `{}` is an empty object. Helm treats this as 'empty', along with `null`, `0`, `""` and `[]`, so `default` and `coalesce` will treat an empty object as value-not-there. I think the merging of different `values.yaml` files happens the same way, but this isn't actually documented properly. However, if any non-empty value is supposed to be an array, it seems clearer to me to specify the default as an empty array rather than an empty object.

Agreed.

amrc-benmorrow (approved):

PR #23, opened 2023-06-30
Add default value for identity.krbKeysOperator.namespaces to values.yaml

AlexGodbehere:


values.yaml
```diff
@@ -24,6 +24,8 @@ identity:
       # @ignore
       pullPolicy: IfNotPresent
   krbKeysOperator:
+    # -- A comma-separated list of namespaces that the KerberosKey Operator should watch for KerberosKey resources in. Defaults to the release namespace if not specified
+    namespaces: {}
```
amrc-benmorrow, line 28: It's a little confusing giving an object here when any non-empty value needs to be a string. I would use `""` or `null` instead.

README.md
```diff
@@ -146,6 +146,7 @@ Production deployment does not differ greatly from development deployment, howev
 | identity.identity.image.registry | string | `"ghcr.io/amrc-factoryplus"` | The registry of the Identity component |
 | identity.identity.image.repository | string | `"acs-identity"` | The repository of the Identity component |
 | identity.identity.image.tag | string | `"latest"` | The tag of the Identity component |
+| identity.krbKeysOperator.image | object | `{"registry":"ghcr.io/amrc-factoryplus","repository":"acs-krb-keys-operator","tag":"latest"}` | A comma-separated list of namespaces that the KerberosKey Operator should watch for KerberosKey resources in. Defaults to the release namespace if not specified |
```
amrc-benmorrow, line 149: ...I don't think this is right.

amrc-benmorrow (changes requested):

amrc-benmorrow (dismissed):

amrc-benmorrow (approved):

PR #27, opened 2023-07-03
Fix some invalid generated YAML.

amrc-benmorrow:
Excessive chomping on Helm templates was causing the YAML divider `---` to end up appended to the end of the previous line.


amrc-benmorrow:
Sorry, this has been merged... it's been squashed and the remote branch has been deleted so it looked like the branch wasn't pushed at all.

PR #30, opened 2023-07-10
Grafana kerberos auth

AlexGodbehere:
- Adds the ability to login to Grafana with Factory+ credentials
- Configures a default datasource in Grafana preconfigured to extract data from the InfluxDB instance using InfluxQL
- Configures an additional datasource in Grafana preconfigured to extract data from the InfluxDB instance using Flux
- Updates the `NOTES.txt` readme to guide users to log into Grafana instead of InfluxDB to query data
- Update InfluxDB Ingester to `v1.0.1`
- Update ACS Auth to `v1.2.0`

**BREAKING CHANGES**
- Update Traefik Helm chart to `23.1.0` and update appropriate CRDs
- Requires fresh Grafana provisioning


amrc-benmorrow (approved):
Is the Traefik upgrade (and associated CRD change) necessary for the plugin to work? If not I would be inclined to split that piece out into a different PR, and consider our upgrade path before merging it.

When we come to release, does this make a breaking change (major version rev) of the ACS helm chart as a whole? I think that probably depends on whether we can get Helm to script the CRD change or not.

PR #38, opened 2023-07-11
Add Git and EDO components

AlexGodbehere:


templates/manager/rbac.yaml
```diff
@@ -82,4 +82,12 @@ rules:
...
-{{- end -}}
\ No newline at end of file
+{{- end}}
+---
+{{- if .Values.edo.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: edge-deployment-operator
```
amrc-benmorrow, line 92: I don't think this ServiceAccount is added to any RoleBindings. The Manager account had a ClusterRoleBinding, but I'm not sure if that was necessary, we might be able to make do with a RoleBinding.

The account needs the same get/create/patch rights as the manager had.

amrc-benmorrow (changes requested):

amrc-benmorrow (approved):

PR #39, opened 2023-07-11
Add Loki as Grafana datasource configuration

AlexGodbehere:
This commit adds a new configuration for Loki as a datasource in Grafana. This will allow us to query and visualize log data from Loki directly in Grafana.


amrc-benmorrow (dismissed):
I have no clue about this :).

PR #40, opened 2023-07-12
Add Switchboard for DNS and TLS automation

AlexGodbehere:
Added Switchboard as a sub-chart to automate DNS and TLS configuration, reducing manual setup. Updated README and template notes to reflect these changes. Hardcoded 'factoryplus-tls' as TLS secret name in the ingress templates and removed 'acs.tlsSecretName' from 'values.yaml' and relevant files. Also removed 'values.original.yaml' as it was redundant. Remember to refer to the Switchboard documentation for more information.


amrc-benmorrow (dismissed):
I have two questions about this:
* It seems like Switchboard is not useful without cert-manager and external-dns. Is the intention to pull those in two?
* This appears to be assuming Switchboard will be able to arrange the issue of a wildcard certificate; all IngressRoutes use the same secret. This may not be possible for some users?

AlexGodbehere:
>  all IngressRoutes use the same secret. This may not be possible for some users?

- [ ] Agreed, these should be separated.

PR #42, opened 2023-07-18
Fix the EDO deployment

amrc-benmorrow:
This gets the git and EDO deployments into a working state.

There is one significant remaining problem: the initial creation of the git repos, and the fact that the repo UUID for the `shared/flux-system` repo needs to be hardcoded into the cluster creation template. This will require changes to the git server code, I think.

This updates the KerberosKey CRD. I don't think we yet have an upgrade strategy here; possibly this will have to be another major release of ACS with (another) manual upgrade step.

### DIFF
--- a/templates/dumps.yaml
+++ b/templates/dumps.yaml
@@ -150,6 +150,7 @@ data:
         {{$classDef_permissionGroup}},
         {{$classDef_wildcard}},
         {{$classDef_serviceAccount}},
+        {{$classDef_serviceRequirement}},
         {{$classDef_cellGateway}},
         {{$classDef_softGateway}}
       ],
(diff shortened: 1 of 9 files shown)


templates/dumps.yaml
```diff
@@ -82,23 +102,32 @@
...
 {{$serviceAccount_dataWarehouse := "388ddbdc-4eb4-4ae8-bbd0-9be32f3c31e8" | quote}}
 {{$serviceAccount_directory := "5cc3b068-938f-4bb2-8ceb-64338a02fbeb" | quote}}
 {{$serviceAccount_manager := "2340e706-1280-420c-84a6-016547b55e95" | quote}}
 {{$serviceAccount_mqtt := "2f42daeb-4521-4522-8e19-85dfb73db88e" | quote}}
-
-# ================================================
-# Generate unique UUIDs for each instance deployed
-# ================================================
-
-{{$user_admin := uuidv4 | quote}}
```
AlexGodbehere, line 98: Don't we want the admin user to have their own UUID in each deployment of ACS?

AlexGodbehere (approved):

PR #43, opened 2023-07-19
Updates to the README.

amrc-benmorrow:
* Document the process of upgrading and ACS install.
* Document that major versions may require manual steps and must not be skipped.
* Document the process of installing from a Git checkout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ Helm is a package manager for Kubernetes that allows you to easily install and m
 ### Install Kubectl
 Kubectl is a command-line tool for controlling Kubernetes clusters. It must be installed on the machine that you'll be using to deploy ACS _from_. To install Kubectl, follow the instructions [here](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
 
+### Install Lens
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,54 +17,171 @@ Helm is a package manager for Kubernetes that allows you to easily install and m
...
 
-To create the `factory-plus` namespace, run the following command:
 ```bash
-kubectl create namespace factory-plus
+helm upgrade acs amrc-connectivity-stack/amrc-connectivity-stack -f $values --version $version
 ```
 
-Finally, install ACS by running the following command.
+specifying the version to upgrade to. You need to use the same `$values`
+file as before, but you don't need to re-specify the namespace.
```
AlexGodbehere, line 153: Is this the case? I'm sure the namespace is required but I could be wrong.

AlexGodbehere (approved):

README.md
amrc-benmorrow: You are definitely wrong :). I use `helm upgrade -f values.yaml acs ..\amrc-connectivity-stack` regularly.

PR #50, opened 2023-08-08
Add visualiser

AlexGodbehere:


AlexGodbehere:
@amrc-benmorrow Is the default broker hard-coded into the `vis` package? If so we'll need to move that to an env var before we package it up and release `v1.0.0`.

amrc-benmorrow:
@AlexGodbehere Yes, it is; and yes, that needs fixing. It's not quite a trivial fix as the whole app is client-side; the server-side piece is currently just dumb file serving. I'll have a look at the best way to do it.

amrc-benmorrow:
This should not be a PR into `v2-stable`, but into `main`. New features should not be going into v2 at this point, only changes that are required to make it actually deployable.

PR #52, opened 2023-08-21
Add switchboard

AlexGodbehere:


amrc-benmorrow (changes requested):
IMHO this should not be going into `v2-stable`. That branch should be reserved for the minimum changes needed to get v2 deployable, and then for important (show-stopper or security) bugfixes as we discover them over the supported life of v2. New features should be going into `main`, ready for v3 when that happens.

AlexGodbehere:
@amrc-benmorrow I disagree - new features are "minor" releases (`v2.*`). This isn't a breaking change so I don't see why it needs to belong in `v3.*`?

PR #73, opened 2023-11-06
[DO NOT MERGE] `main` to `v2-stable` comparison

AlexGodbehere:
This PR will never be merged - it simply provides a place to have a discussion about differences between the current state of `v2-stable` and `main`.

**DO NOT MERGE**


crds/kerberos-keys-crd.yaml
AlexGodbehere, line 1: @amrc-benmorrow is this CRD definition up to date or wildly outdated?

templates/dumps.yaml
```diff
@@ -1,6 +1,3 @@
-# Only run this if the ConfigMaps don't already exist
-{{- if and (not (lookup "v1" "ConfigMap" .Release.Namespace "configdb-json-dumps")) (not (lookup "v1" "ConfigMap" .Release.Namespace "auth-json-dumps")) }}
-
```
AlexGodbehere, line 3: @amrc-benmorrow Any idea why this was removed? This whole thing is a candidate for `acs-service-setup`.

crds/kerberos-keys-crd.yaml
amrc-benmorrow: It's correct for the version of krbkeys deployed by ACS. The new version has an updated CRD which lives [here](https://github.com/AMRC-FactoryPlus/acs-krb-keys-operator/blob/main/crd/kerberos-keys.yaml) but that wants pulling into ACS at the same time as the new krbkeys operator. In general whenever we pull in a new krbkeys release the CRD should be updated.

templates/manager/rbac.yaml
```diff
@@ -1,10 +1,50 @@
...
+  name: krb-keys-manager
+rules:
+  -   apiGroups: [factoryplus.app.amrc.co.uk]
+      resources: [kerberos-keys]
+      verbs: [list, get, create, update, delete, watch, patch]
+  -   apiGroups: [factoryplus.app.amrc.co.uk]
+      resources: [kerberos-keys/status]
+      verbs: [list, get, watch]
+---
+
```
AlexGodbehere, line 46: Am I correct in thinking that the Manager no longer needs these?

crds/kerberos-keys-crd.yaml
AlexGodbehere: Ok, I'll leave this for you to coordinate. This `main` branch is ready for any changes to be made.

templates/dumps.yaml
amrc-benmorrow: https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/commit/5183992396a9b0af112147b5396337feafe2b0c5

amrc-benmorrow: I was having problems with them not getting updated when running `helm upgrade`, because they already existed and so Helm was skipping the whole file.

amrc-benmorrow: I think the account UUID generation should be able to be handled by krbkeys now. If we create appropriate (fixed UUID) Service Requirement groups, to which permissions are granted in the dump file (or by service-setup), then krbkeys can now create fresh (per-deployment UUID) accounts and place them in the groups.

Issue #89, opened 2023-12-12
Allow all containers to be pulled from private registries


derme302:
Currently most of the containers in the stack can be overridden to be pulled from a private registry, except for `kubegres-controller-manager`.

It would be great if the registry address for both `kubegres-controller-manager` and `manager` were exposed in the `values.yaml`. Along with this it would be great to have `imagePullSecrets` exposed as these exist in a different namespace to the rest of the containers in the stack.

amrc-benmorrow:
I'd be interested to know what your use case is here... we are currently looking into running ACS on edge systems isolated from the Internet, and we have had success using [Harbor](https://goharbor.io/) as a pull-through cache along with using k3s' [private registries configuration](https://docs.k3s.io/installation/private-registry) to configure machines to pull images through the cache. But obviously this requires some part of the network has Internet access.

derme302:
The use case is a factory environment with no internet connection. Our images are currently in a private registry and require authentication to retrieve. 

If we wanted to remove the need to authenticate we'd need to setup another registry with just the Factory+ containers. K3's private registry configuration would work but we don't have the option to make modifications to the cluster at this point.

amrc-benmorrow:
I believe this is now resolved with #98, #100, #113 and #114. If there are still issues please feel free to reopen.

derme302:
Awesome, thanks @amrc-benmorrow!

PR #97, opened 2024-01-09
Expose all images and add imagePullSecrets

derme302:
This is a potential solution to #89 it makes two changes:

1. Adds the ability to define `imagePullSecrets` for images that are in private registries
2. Exposes some of the images that were previously hardcoded

It also changes the version of the `postgres` container used by the manager service from `14.3` to `14.10`, but that could be changed if needed. It just seems to make sense to use the same version for both containers.

Run in an isolated cluster and seems to all be functioning with no errors.

### DIFF
--- a/templates/auth/auth.yaml
+++ b/templates/auth/auth.yaml
@@ -17,6 +17,10 @@ spec:
         component: auth
         factory-plus.service: auth
     spec:
+      {{- with .Values.acs.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       volumes:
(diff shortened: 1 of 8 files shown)


amrc-benmorrow:
@derme302 I have reworked the changes in here on top of main, and backported to v2-stable as #100 . Are you able to check those changes still work in your environment?

derme302:
Thanks @amrc-benmorrow for adding that in so quickly. I've tested #100 but I found some auth issues with my registry, which I included in #101

amrc-benmorrow:
I believe this has now been fully resolved with #98, #100, #113 and #114.

PR #101, opened 2024-01-11
Fix missing imagePullSecrets

derme302:
When I tried to deploy the `backport/images` branch I ran into some authentication errors. This is with `pullPolicy: Always`  on every image in my `values.yaml` file.

I'm a bit unsure about two things:
- Why this wasn't an issue when I tested my branch
- Why `imagePullSecrets` isn't needed in every yaml file that references a container

With these changes I can successful deploy to my cluster.


amrc-benmorrow:
Hmm, I don't understand either. I would expect the pull secrets to be required everywhere. I would have said 'maybe some of the templates don't honour `pullPolicy`' but for example `cmdesc` does and that doesn't currently have the `imagePullSecrets`. I will need to go through and add the secrets to every Pod, I think.

derme302:
Hey @amrc-benmorrow, I've gone through all the pods and added `imagePullSecrets` to all of them. I've pulled the changes from `v2-stable` and tested them but I'm not sure that means much as the original PR worked for me.

PR #110, opened 2024-01-23
Central service support for edge clusters

amrc-benmorrow:
Configuring the services for edge cluster support requires a large number of object UUIDs which are specific to this installation: git repo UUIDs, Helm template UUIDs, UUIDs for the groups of edge accounts created automatically.

Attempting to dynamically generate UUIDs in the Helm chart did not work very well. Instead, move more of this work into service-setup.

* Set up the configuration needed for the edge cluster infrastructure.
* Set up the configuration needed for the Manager to create edge clusters and deployments.
* Start moving setup from `dumps.yaml` into service-setup. This process is very far from complete, and there will be bootstrapping questions to answer before it is finished.


templates/clusters/clusters.yaml
AlexGodbehere, line 1: Didn't we agree on `cluster-manager` as the name and `clusters.*` as the endpoint?

amrc-benmorrow: We'll do this as another PR

AlexGodbehere (approved):

PR #144, opened 2024-03-26
Use chart version for images

AlexGodbehere:
Use the helper function to use the chart version for the image tags if not specified in the values file.

- Replaced all remaining static calls to images with the helper functions.

### DIFF
--- a/deploy/templates/_helpers.tpl
+++ b/deploy/templates/_helpers.tpl
@@ -39,7 +39,8 @@ Create chart name and version as used by the chart label.
 Define the image for a container.
 */}}
 {{- define "amrc-connectivity-stack.image-name" -}}
-{{- .image.registry }}/{{ .image.repository }}:{{ .image.tag }}
+{{- $defaultTag := default .Release.Version .image.tag }}
+{{- .image.registry }}/{{ .image.repository }}:{{ $defaultTag }}
 {{- end }}
 
 {{- define "amrc-connectivity-stack.image" -}}
--- a/deploy/templates/_helpers.tpl
+++ b/deploy/templates/_helpers.tpl
@@ -39,7 +39,7 @@ Create chart name and version as used by the chart label.
 Define the image for a container.
 */}}
 {{- define "amrc-connectivity-stack.image-name" -}}
-{{- $defaultTag := default .Release.Version .image.tag }}
+{{- $defaultTag := coalesce .image.tag $.Values.acs.defaultTag $.Release.Version }}
 {{- .image.registry }}/{{ .image.repository }}:{{ $defaultTag }}
 {{- end }}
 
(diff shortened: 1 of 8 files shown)


deploy/templates/identity/kdc.yaml
```diff
@@ -48,7 +48,7 @@ spec:
         # Do as little work as possible as root... We need to chown the storage
         # as the default permissions will not be correct.
         - name: chown-storage
-          image: "{{ .Values.identity.identity.image.registry }}/{{ .Values.identity.identity.image.repository }}:{{ .Values.identity.identity.image.tag }}"
+          image: "{{ include "amrc-connectivity-stack.image-name" .Values.identity.identity }}"
```
amrc-benmorrow, line 51: Is there a reason these aren't using `amrc-connectivity-stack.image`, which includes `imagePullPolicy` from the values file?

deploy/templates/manager/manager.yaml
```diff
@@ -38,7 +38,7 @@ spec:
         # Do as little work as possible as root... We need to chown the storage
         # as the default permissions will not be correct.
         - name: chown-storage
-          image: "{{ .Values.manager.image.registry }}/{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag }}-backend"
+          image: "{{ include "amrc-connectivity-stack.image-name" .Values.configdb }}-backend"
```
amrc-benmorrow, line 41: `configdb`...

deploy/templates/manager/schema-import-cron.yaml
```diff
@@ -29,7 +29,7 @@ spec:
                 name: krb5-conf
           containers:
             - name: cron
-              image: "{{ .Values.manager.image.registry }}/{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag }}-backend"
+              image: "{{ include "amrc-connectivity-stack.image-name" .Values.configdb }}-backend"
```
amrc-benmorrow, line 32: ...`configdb` to here

deploy/templates/_helpers.tpl
```diff
@@ -39,7 +39,8 @@ Create chart name and version as used by the chart label.
 Define the image for a container.
 */}}
 {{- define "amrc-connectivity-stack.image-name" -}}
-{{- .image.registry }}/{{ .image.repository }}:{{ .image.tag }}
+{{- $defaultTag := default .Release.Version .image.tag }}
```
amrc-benmorrow, line 42: I think this is still scoped within the `.Values.*` passed in as context, so `.Release.Version` won't work. We need `$.Release.Version`.

I think it's also worth making this `coalesce .image.tag $.acs.defaultTag $.Release.Version` or something to allow a global override of the image tag. In particular I think a Flux HelmRelease pulling directly from a GitRepository results in a weird `.Release.Version` which won't work correctly.

amrc-benmorrow (changes requested):

deploy/templates/identity/kdc.yaml
AlexGodbehere: No. I needed to use the `image-name` one to include the suffix for the manager and it slipped through with them. I'll update now.

amrc-benmorrow (approved):

amrc-benmorrow (approved):

PR #237, opened 2024-04-10
Activity monitor

AlexGodbehere:
The Activity Monitor is an alternate view of activity within your ACS infrastructure. It displays live device/node activity, the number of messages sent by each device/node and indicates BIRTH, DEATH and NCMD messages via colour changes.

It can be accessed at `http(s)://activity.*`, where `*` is the base URL of your ACS installation.

2024-04-10 at 08 21 05


AlexGodbehere:
@amrc-benmorrow would it make more sense to update the service client to remove the blank space after await instead of patching it?

Issue #235, opened 2024-04-03
[directory] Clear old Alerts


amrc-benmorrow:
If we see a birth certificate from a device and it does not contain a given Alert, that Alert no longer exists. It should be cleared from the `/alert` API, at least.

Whether we should remove all record of it is a different question. Possibly we should record all schema instance UUIDs in the ConfigDB for conflict detection?

Issue #184, opened 2024-02-07
[directory] Alerts API doesn't include Links


amrc-benmorrow:
The Links associated with an Alert are quite important, and we're not indexing them yet. For example, the Monitor alerts use Links to show which Node they are monitoring.

Perhaps we need a more general Links index?

PR #238, opened 2024-04-17
Record Links in the Directory

amrc-benmorrow:
* Parse Links out of Sparkplug packets and record them in the Directory.
* Make change-notify available for Links.
* Make the Links for an Alert available via the Alerts API.
* Remove stale Links and Alerts from the API.

Closes: #235, #184


AlexGodbehere:
We'll also need to update the API specification in the portal if we've added endpoints to the Directory.

PR #257, opened 2024-04-22
Lint everything

amrc-benmorrow:
* Create a `make lint` top-level Makefile target which runs through all subdirs and lints anything which defines a lint rule.
* Create a Makefile which can lint the `deploy` subdir.
* Use this in the GH Action.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,4 +1,4 @@
-name: Helm Lint
+name: Lint everything
 
 on:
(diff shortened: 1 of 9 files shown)


amrc-benmorrow:
This needs rebasing before it can be considered for merge.

amrc-benmorrow:
This is too far out of date to be useful. We have a top-level `lint` target now; wiring it up to GHA won't be difficult if we want to.

PR #290, opened 2024-06-28
Alerts interface

AlexGodbehere:
This update brings a new administration interface for ACS, a place where most human-facing operations will be performed going forward. For now it includes a new version of the activity monitor and a UI to view alerts raised across the infrastructure.

Google Chrome 2024-06-28 14 23 14
Google Chrome 2024-06-28 14 23 46


AlexGodbehere:
In the process of trying to get build vars in at runtime using https://import-meta-env.org/guide/getting-started/introduction.html and ran out of time.

This takes a `.env` file in the Dockerfile `CMD`, but we don't have one. Have the Helm chart supply one instead of environment variables in the traditional sense.

PR #316, opened 2024-07-22
Support external Edge Agent drivers

amrc-benmorrow:
* Define a protocol, running over MQTT, for communication between Edge Agent and drivers.
* Set up the Edge Agent to run a local broker for drivers to connect to.
* Remove the Device subclasses; move any functionality into the DeviceConnections.
* Create a Driver DeviceConnection communicating with the drivers.
* Implement testing and modbus drivers.
* Make krbkeys able to create local secrets so drivers can authenticate to the Edge Agent.
* Update the edge helm charts to deploy an Edge Agent with drivers.

There is currently no Manager support for drivers. This will need more work.


acs-edge/lib/device.ts
```diff
@@ -144,7 +171,7 @@ export abstract class Device {
     _defaultMetrics: sparkplugMetric[]              // The default metrics common to all devices
     #isAlive: boolean                               // Whether this device is alive or not
     _isConnected: boolean                               // Whether this device is ready to publish or not
-    #deathTimer: ReturnType<typeof setTimeout>      // A "dead mans handle" or "watchdog" timer which triggers a DDEATH
+    //#deathTimer: ReturnType<typeof setTimeout>      // A "dead mans handle" or "watchdog" timer which triggers a DDEATH
```
AlexGodbehere, line 174: What's the reason for disabling this?

acs-edge/lib/devices/driver.ts
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Stop a previously registered subscription for  metric changes.
+     * @param deviceId The device ID we are cancelling the subscription for
+     * @param stopSubCallback A function to call once the subscription has been cancelled
+     */
+    //async stopSubscription(deviceId: string, stopSubCallback: Function) {
+    //}
+    
+    #newTopic () {
+        while (true) {
```
AlexGodbehere, line 127: What does this do, and why?

AlexGodbehere (approved):

acs-edge/lib/device.ts
amrc-benmorrow: > I have disabled the watchdog timeout; it fires much too quickly when
> you're pretending to be a device manually. I'm not sure a fixed timeout
> with no ping is the right thing to do anyway; there are valid reasons
> why a connection might not speak for 10s.

acs-edge/lib/devices/driver.ts
amrc-benmorrow: The edge driver protocol maps addresses (which might be long strings and unsuitable for MQTT topic components) into data topic names assigned by the edge agent. This generates a new random number to use as a data topic name, and checks it actually is new before returning it.

PR #341, opened 2024-09-06
Unified Namespace

AlexGodbehere:
This PR adds support for the concept of UNS Ingesters, which take channels of data (in this case, Sparkplug), and publishes the human-readable content to `UNS/v1`. This PR includes the required permissions, pricipals and changes to the values file.

It also includes a new UNS Historian (`historian-uns`) to persist the UNS topics to InfluxDB and renames the `sparkplug-influxdb-ingester` to `historian-sparkplug`.


acs-service-setup/dumps/uns.yaml
```diff
@@ -0,0 +1,45 @@
...
+    !u ACS.ServiceAccount.HistorianUNS: { name: "Historian (UNS)" }
+  !u UUIDs.App.MQTTPermissionTemplate:
+    !u ACS.Perm.MQTT.WriteToEntireUNS: "UNS/v1/#: w"
+    !u ACS.Perm.MQTT.ReadEntireUNS: "UNS/v1/#: rs"
+---
+service: !u UUIDs.Service.Authentication
+version: 1
+principals:
+  - uuid: !u ACS.ServiceAccount.SparkplugIngester
+    kerberos: sv1sparkplugingester@{{realm}}
```
AlexGodbehere, line 30: @amrc-benmorrow how do we reference the realm here? I thought it might get templated but it doesn't seem to (image below is from deployment of `v3.3.0-dev.2`.

image

amrc-benmorrow: You don't, you let krbkeys create these entries. These service account UUIDs should not be fixed in any case, they should be per-installation.
* Create an Auth group _Sparkplug Ingesters_. This does have a fixed UUID.
* Grant that group appropriate permissions. If all it needs is _Global Debuggers_ then you don't need a new group but I think it will need UNS publish permission as well? You should be able to make the _Sparkplug Ingesters_ group a member of the _Global Debuggers_ group, which grants read-and-rebirth.
* Request account creation on the `sv1sparkplugingester` KerberosKey. The [central Monitor](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/blob/45407a96f5ffb65fe6e0441064fb9c37809438d8/deploy/templates/auth/principals/service-clients.yaml#L116) is the only account to use this pattern so far but it's how I'd like to do things going forward.

uns-ingester-sparkplug/bin/ingester.js
amrc-benmorrow, line 1: Is this file even used? What's the relationship between this and the `.ts`?

uns-ingester-sparkplug/lib/mqttclient.ts
```diff
@@ -0,0 +1,522 @@
...
+                    logger.info(`✨ Device ${topic.address.group}/${topic.address.node}/${topic.address.device} is unknown, requesting birth certificate`);
+
+                    // Create debounce timout for this device
+                    this.setNestedValue(this.birthDebounce, [topic.address.group, topic.address.node, topic.address.device], true);
+                    setTimeout(() => {
+                        delete this.birthDebounce?.[topic.address.group]?.[topic.address.node]?.[topic.address.device];
+                    }, Math.floor(Math.random() * (10000 - 5000 + 1) + 5000));
+
+                    // Request birth certificate
+                    let response = await this.serviceClient.fetch({
```
amrc-benmorrow, line 337: `await this.serviceClient.CmdEsc.rebirth(topic.address)`

amrc-benmorrow (changes requested):
I have to ask: is it really a good idea to record everything into Influx again from the UNS? Would it not be better just to attach the ISA95 tags to the data from within the ordinary Sparkplug ingester?

acs-service-setup/dumps/uns.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - !u ACS.Group.HistorianUNS
+configs:
+  !u UUIDs.App.Info:
+    !u ACS.Perm.MQTT.WriteToEntireUNS: { name: "MQTT: Write to entire UNS" }
+    !u ACS.Perm.MQTT.ReadEntireUNS: { name: "MQTT: Read entire UNS" }
+    !u ACS.Group.SparkplugIngesters: { name: "Sparkplug Ingesters" }
+    !u ACS.Group.HistorianUNS: { name: "Historian UNS" }
+  !u UUIDs.App.MQTTPermissionTemplate:
+    !u ACS.Perm.MQTT.WriteToEntireUNS: "UNS/v1/#: w"
+    !u ACS.Perm.MQTT.ReadEntireUNS: "UNS/v1/#: rs"
```
amrc-benmorrow, line 24: This will insert a string as the entire ConfigDB entry. Are these supposed to be sub-objects? 
`{ "UNS/v1/#": "w" }` or
```
    !u ACS.Perm.MQTT.ReadEntireUNS:
        UNS/v1/#: rs
```

acs-service-setup/lib/uuids.js
```diff
@@ -52,9 +57,14 @@ export const ACS = {
     Service: {
         Manager:                "619eecab-742d-4824-8b97-bcae472e5c04",
     },
+    ServiceAccount: {
+        SparkplugIngester:      "e5939141-a3d5-4ab8-b609-b450ca785d7c",
+        HistorianUNS:           "b6d0cb6a-708b-424f-a062-8c30e4b2fc4c",
+    },
```
amrc-benmorrow, line 63: I think these are now redundant?

deploy/templates/auth/principals/service-clients.yaml
```diff
@@ -122,3 +122,47 @@ spec:
...
+  principal: sv1sparkplugingester@{{ .Values.identity.realm | required "values.identity.realm is required!" }}
+  secret: uns-ingester-sparkplug-keytabs/client
+  account:
+    class: e463b4ae-a322-46cc-8976-4ba76838e908
+    name: Sparkplug Ingester
+    groups:
+      - e414d355-b991-429b-8f5d-97e823ff71f5
+    sparkplug:
+      group: {{ .Values.acs.organisation }}-Service-Core
+      node: SparkplugIngester
```
amrc-benmorrow, line 143: A prinicipal only needs a Sparkplug Node identity if it will be publishing to Sparkplug. I don't think the ingester does that?A Sparkplug identity is not needed just to consume Sparkplug data.

deploy/templates/auth/principals/service-clients.yaml
```diff
@@ -122,3 +122,47 @@ spec:
...
+  principal: sv1historianuns@{{ .Values.identity.realm | required "values.identity.realm is required!" }}
+  secret: historian-uns-keytabs/client
+  account:
+    class: e463b4ae-a322-46cc-8976-4ba76838e908
+    name: Historian UNS
+    groups:
+      - 03f5f08a-f61e-4134-8f66-b2951e3bbb69
+    sparkplug:
+      group: {{ .Values.acs.organisation }}-Service-Core
+      node: HistorianUNS
```
amrc-benmorrow, line 166: A historian also probably doesn't publish (unless it's reporting on its own status?)

deploy/templates/auth/principals/service-clients.yaml
```diff
@@ -122,3 +122,47 @@ spec:
...
+      - e414d355-b991-429b-8f5d-97e823ff71f5
+    sparkplug:
+      group: {{ .Values.acs.organisation }}-Service-Core
+      node: SparkplugIngester
+
+{{- end }}
+---
+# In the future if we have any more ingesters then this should be
+# enabled if any of the ingesters are enabled.
+{{- if .Values.unsIngesters.sparkplug.enabled }}
```
amrc-benmorrow, line 149: You already have `historians.uns`. The UNS republishing can be useful as an alternative to needing to read Sparkplug for live data without duplicating the data in the historian.

deploy/values.yaml
```diff
@@ -202,11 +215,24 @@ cmdesc:
   # -- Possible values are either 1 to enable all possible debugging, or a comma-separated list of debug tags (the tags printed before the log lines). No logging is specified as an empty string.
   verbosity: 1
 
-warehouse:
-  # -- Whether or not to enable the Warehouse component
-  ingester:
+historians:
+  uns:
     enabled: true
```
amrc-benmorrow, line 220: I would consider carefully whether this should default to enabled. I think this will effectively double the historian storage requirement.

amrc-benmorrow (changes requested):

acs-service-setup/dumps/uns.yaml
djnewbould: yep, these should be sub-objects.

deploy/values.yaml
djnewbould: I did notice this when testing. I think it should be disabled by default.

deploy/templates/auth/principals/service-clients.yaml
djnewbould: I agree, I've changed it to use the historian uns value.

amrc-benmorrow (dismissed):

PR #342, opened 2024-09-16
Deploy schemas via service-setup

amrc-benmorrow:
* Refactor the service-setup Job to allow an arbitrary list of containers to be run under the control of the Helm values file.
* Include a new step to deploy the schemas to the ConfigDB.

### DIFF
--- a/deploy/values.yaml
+++ b/deploy/values.yaml
@@ -115,7 +115,7 @@ serviceSetup:
   # an array it can only be overridden as a whole, so the step
   # definitions are separate so they can be added to rather than
   # replaced.
-  runSteps: [service-setup, edge-helm]
+  runSteps: [service-setup, acs-schemas, edge-helm]
   # This defines what to run for each step.
   steps:
     service-setup:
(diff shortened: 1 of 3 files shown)


deploy/values.yaml
```diff
@@ -111,32 +111,48 @@ monitor:
...
+          repoMirror:
+            helm:
+              name: Edge Helm charts
+              pull: {}
+          # Helm charts to deploy to the edge; these default to the charts
+          # created automatically but can be overridden to customise
+          helmChart:
+            # Chart to deploy an edge cluster
+            #cluster: null
+    acs-schemas:
```
AlexGodbehere, line 143: Wouldn't it make more sense to add this as a fixup in `service-setup`? Is there a reason for wanting to break this out at this level?

AlexGodbehere (approved):

deploy/values.yaml
amrc-benmorrow: If you mean 'why is it another container' then: because the schemas are not in the monorepo. So we will have e.g. `acs-schemas:v1.0.0` which is not directly connected to the ACS versioning. This may be the wrong decision, it may be better to move them into the monorepo, but using different schemas seems like a likely thing for people to want to do. (Also using multiple schema-loading images, e.g. ACS and also some private schemas.)

amrc-benmorrow: We are going to move the schemas into the monorepo, however it will still be best to keep them as a separate image so people can choose what to deploy.

PR #366, opened 2024-10-28
Implement a WS notify interface on the ConfigDB

amrc-benmorrow:
Specify a new change-notify interface. This abandons the idea of trying to send notifications over Sparkplug in favour of using a direct WebSocket connection to each service. Hopefully this will be more robust and will avoid the bootstrapping issues that turn up when we try to implement change-notify on the Auth service. This new interface uses the `notify/v2` path prefix to avoid conflict with the (probably now abandoned) `notify/v1` spec which is on the dyn-deploy branch.

The new interface runs entirely over a WebSocket connection, which is used to multiplex multiple subscriptions to sources of notifications. Each subscription returns a sequence of HTTP responses, equivalent to the client polling the REST interface and ignoring duplicate responses. This means that error responses such as 404 and 403 can be returned when appropriate while allowing a new 200 response to be returned later. The only source of notifications currently defined is a `WATCH` of a particular URL, which returns a new response whenever the resource at that URL changes.

Implement the new interface for the ConfigDB. The current implementation only supports `WATCH` on the `app/:app/object/:object` endpoint, meaning a client can watch for changes to a particular config entry. The interface will need extension to be useful but this PR gets the framework in place.

### DIFF
--- a/acs-configdb/Makefile
+++ b/acs-configdb/Makefile
@@ -1,6 +1,8 @@
 top=..
 include ${top}/mk/acs.init.mk
 
+eslint=	bin lib
+
 repo?=acs-configdb
 k8s.deployment?=configdb
 
--- a/acs-configdb/bin/api.js
+++ b/acs-configdb/bin/api.js
@@ -9,8 +9,8 @@
 import url from "url";
 import express from "express";
 
-import { Debug, ServiceClient, UUIDs }  from "@amrc-factoryplus/service-client";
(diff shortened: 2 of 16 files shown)


djnewbould (approved):
Looks like a good addition. The added documentation provides a good high level understanding of the interface but I think the addition of some JSDoc function comments on the code could be helpful for maintainers too.

PR #368, opened 2024-11-01
Implement SEARCH notifications for the ConfigDB

amrc-benmorrow:
Specify a new change-notify request, SEARCH. This allows a client to request notifications for changes to all URLs under a given path. The intended use-case is for watching `app/:app/object/` to track updates to configs we are interested in.

Some apps (notably the edge-sync operator) will filter out many objects. Support this in the API by allowing the client to apply a filter to the notifications. Filtered-out objects are not returned and updates to them are ignored, however the client is kept informed when objects enter or leave the set of interest.


docs/notify-v2.md
djnewbould, line 1: The documentation provided here is very detailed.

djnewbould (dismissed):

PR #375, opened 2024-11-14
New Auth spec

AlexGodbehere:
Specify the data model used for the new Auth service. This defines ACLs in terms of templates, allowing a single grant to expand to multiple related grants on different services. The template language is able to look up information from other services allowing support for dynamically-defined ACLs.

### DIFF
--- a/docs/auth/sexpr.md
+++ b/docs/auth/sexpr.md
@@ -1,0 +1,377 @@
+# S-expression based permission templates
+
+This is a proposal for an Auth service redesign based on s-expressions
+expressed as JSON arrays.
+
+* Groups can only be used for principals or targets (nouns). Groups of
+  permissions are no longer expanded.
+
+* The function `members` expands the members of a group. Expanding a
+  UUID which does not represent a group returns a singleton list.
+
+* The principal of an ACE is always a UUID and is always expanded with
+  `members`. The ACE is granted to any identifier belonging to a UUID on
+  the resulting list. Principal UUIDs should not normally be used as
+  group UUIDs.
+
+* The target of an ACE is a JSON value, which must be either an object,
+  a string or `null`. Arrays are reserved to represent template calls.
+  In some cases this will be a string representing a UUID but this is up
+  to the service to interpret. Wildcards and group expansion are also up
+  to the service.
+
+* A permission is always a UUID. There are two categories of permission:
+  base permissions which are returned to the requesting service, and
+  permission templates. Templates are defined in the Auth service and
+  permissions defined as templates will never be returned to a service.
+
+## Permission templates
+
+A permission template is a function which accepts the principal UUID and
+the target value and returns a list of ACEs. When ACLs are looked up all
+permission templates are expanded recursively and the result included in
+the list of ACEs returned. Some form of recursion control will obviously
+be necessary.
+
+A template definition is represented in JSON as an s-expression built
+from arrays and strings.
+
+* A 'base value' is `null`, a string, or an object whose values are all
+  base values.
+
+* An array represents a function call. The first member of the array
+  names the function to call. This must be a string and it names a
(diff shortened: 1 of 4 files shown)


docs/auth/sexpr.md
```diff
@@ -0,0 +1,563 @@
...
+
+    # Template definitions
+    ReadOwnConfig → [[app]
+      [ReadConfig: { app: [app], obj: [principal] }]]
+
+    SpTopic → [[type addr]
+      [if [has [addr] device] 
+        [format "spBv1.0/%s/N%s/%s" [addr group] [type] [addr node]]
+        [format "spBv1.0/%s/D%s/%s" 
+          [addr group] [type] [addr node] [addr device]]]]
```
AlexGodbehere, line 292: Why does this have four arguments but only three `%s`?

docs/auth/sexpr.md
```diff
@@ -0,0 +1,563 @@
...
+  value. (List)
+
+* `let` (*): `["let" bindings ...body]`. The `bindings` must be a list
+  with an even number of items; odd numbered items must be strings and
+  name new bindings. Each is followed by a value. A new set of bindings
+  is constructed starting from the current set. The first value is
+  evaluated using the current bindings and a binding for the first name
+  is added with the result. Each subsequent value is evaluated using the
+  new bindings and added to the set. Finally `body` is list-evaluated
+  using the new bindings and the result returned. (List)
```
AlexGodbehere, line 207: Can you explain this in different terms? I don't entirely understand and as such can't work out what line 294 is doing below.

docs/auth/sexpr.md
```diff
@@ -0,0 +1,563 @@
...
+generalised into a template language defined in S-expressions (LISP data
+structures) expressed as JSON arrays.
+
+## Data structure
+
+### Identity and ownership
+
+* Sparkplug Nodes are principals; their Sparkplug address information is
+  moved into the Auth service to sit alongside the Kerberos identity as
+  an alternative name.
```
AlexGodbehere, line 19: Can you elaborate on what this means? Does this contradict the UNS direction, should we choose to pursue it? With the UNS concept Sparkplug is only one of potentially many channels of data and we could expect other top-level namespaces (and their ingester pods) sending data into the `/UNS/v1` namespace.

docs/auth/sexpr.md
amrc-benmorrow: This is 'variable' assignment (scare quotes because they aren't actually variable). It's equivalent to a block in Javascript with a set of `const` assignments at the top. Given `[let [a VALUE] ...]` the `...` is evaluated, and anywhere inside we replace `[a]` with `VALUE`. So:
```
[let [a 42] [a]] -> [42]
[let [a 42 b "abc"] [b] [a]] -> ["abc" 42]
[let [a 42 b "abc"] [format "%s/%s" [a] [b]]] -> ["42/abc"]
```
In the example on L294, `node` is a variable created for the scope of the let. `[id [principal] sparkplug]` is the value assigned to that variable, which evaluates to an object like `{ group: "Core", node: "Node" }` giving our assigned Sparkplug address. The body of the let contains a map, which iterates over a list built using the `node` variable we assigned to; the map creates another variable `addr`, and the expansion below shows the various stages in evaluating it.

amrc-benmorrow: I suppose given this is JSON we could actually use the syntax `[let {a: 42} [a]]`. I wonder if that would be less confusing?

docs/auth/sexpr.md
amrc-benmorrow: Looking at this again, it's completely wrong. The correct version is here: https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/blob/5774de3a7f56ba603689b3cec970cbc59405124a/acs-auth/bin/sexpr.js#L218

PR #377, opened 2024-11-21
Apply CoreDNS patch by default

AlexGodbehere:
I frequently find myself applying this, especially with air-gap installs. @amrc-benmorrow what do you think and can you see any issues with including this?


amrc-benmorrow:
I don't see any issue with this, in principle.

Is there a piece missing? How does CoreDNS know to read this ConfigMap?

amrc-benmorrow:
This has been implemented.

PR #379, opened 2024-11-27
Create an edge MQTT broker helm chart

amrc-benmorrow:
Create a new edge helm chart which deploys an MQTT broker. This broker is exposed to the cluster as service; the chart also provides the option to expose it on an external IP.

No security is currently implemented, beyond a default of `internalTrafficPolicy: Local` which should prevent off-node pods from contacting the broker but is not intended as a security feature.

### DIFF
--- a/edge-helm-charts/charts/mqtt-broker/templates/service.yaml
+++ b/edge-helm-charts/charts/mqtt-broker/templates/service.yaml
@@ -1,0 +1,18 @@
+{{- $k8sname := include "acs.k8sname" . }}
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ .Chart.Name }}-{{ $k8sname }}
+spec:
+  selector:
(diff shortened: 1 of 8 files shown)


edge-helm-charts/charts/mqtt-broker/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
+{{- $k8sname := include "acs.k8sname" . }}
+apiVersion: v1
+kind: Service
```
AlexGodbehere, line 3: Does this default to `NodePort` type?

amrc-benmorrow: ClusterIP. The external access is provided by `externalIPs` and doesn't need a NodePort. A NodePort would be restricted to a stupid port number; with `externalIPs` we just expose the Service port (1883) directly on an external IP.

AlexGodbehere (approved):

PR #381, opened 2024-11-27
Fix UNS historian null reference

djnewbould:
If an MQTT packet was received without custom properties, an exception was thrown. The added optional chaining prevents this error.


amrc-benmorrow (approved):
I've rebased onto main.

PR #391, opened 2025-01-14
Fix finalizers preventing namespace deletion

djnewbould:
Finalizers on custom resource definitions cause both the factory-plus namespace to be stuck in a terminating state. The added job runs on the helm post delete hook and patches the Kerberos Key to remove the finalizers. 
- [x] Fix factory-plus namespace deletion.

## How to test 
1. Run `Helm uninstall acs -n factory-plus`.
2. Wait for the job to complete.
3. Run `kubectl delete namesapce factory-plus` to delete the factory-plus namespace.
4. The namespace should delete and not be stuck in a terminating state.

### DIFF
--- a/deploy/templates/hooks/post-delete.yaml
+++ b/deploy/templates/hooks/post-delete.yaml
@@ -1,0 +1,71 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: post-delete-cleanup
+  annotations:
+    "helm.sh/hook": post-delete
+spec:
+  ttlSecondsAfterFinished: 30
+  template:
+    spec:
+      serviceAccountName: {{ .Release.Namespace }}-cleanup
+      containers:
+        - name: script-runner
+          image: bitnami/kubectl:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
+              # Your shell script starts here
+              echo "Starting cleanup..."
+              for i in $(kubectl -n {{ .Release.Namespace }} get kerberos-keys -o name); do kubectl -n {{ .Release.Namespace }} patch $i --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'; done
+              echo "Cleanup complete!"
+      restartPolicy: Never
+  backoffLimit: 3
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Namespace }}-cleanup
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-delete,post-delete
+    "helm.sh/hook-delete-policy": before-hook-creation
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Namespace }}-cleanup-role
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-delete,post-delete
+    "helm.sh/hook-delete-policy": before-hook-creation
+
+rules:
+  - apiGroups: [ "factoryplus.app.amrc.co.uk" ]
+    resources: [ "kerberos-keys" ]
+    verbs: [ "get", "list", "patch" ] # Specify only necessary actions
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ .Release.Namespace }}-cleanup-binding
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-delete,post-delete
+    "helm.sh/hook-delete-policy": before-hook-creation
+
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Namespace }}-cleanup
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: {{ .Release.Namespace }}-cleanup-role
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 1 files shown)


deploy/templates/hooks/post-delete.yaml
```diff
@@ -0,0 +1,71 @@
...
+              echo "Cleanup complete!"
+      restartPolicy: Never
+  backoffLimit: 3
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Namespace }}-cleanup
```
amrc-benmorrow, line 31: I'm not sure there's any need for this double namespacing? The resource is already in `Release.Namespace`.

djnewbould: I've changed this in my latest commit.

djnewbould:
> This looks good. Getting an edge cluster to uninstall cleanly will be less straightforward as there are multiple charts involved, but making the `edge-cluster` chart uninstall on its own will be a good start.
> 
> I presume you've tested that Helm is willing to run a `post-delete` hook with the kerberos-keys still present? I wasn't sure it would.
> 
> When it comes to the LocalSecret resources you may run into an issue as this CRD is not yet installed onto the central cluster. Probably the best way forward is to make sure it is; copy the CRD from `acs-krb-keys-operator/crd` to `deploy/crds`.

I didn't notice any issues with the local secret resource,. Do you think its worth copying across anyway?

edge-helm-charts/charts/edge-cluster/templates/post-delete.yaml
```diff
@@ -0,0 +1,75 @@
...
+          image: bitnami/kubectl:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
+              NAMESPACE={{ .Release.Namespace }}
+              echo "Starting cleanup..."
+              # Set the namespace to clean up
+              NAMESPACE=fplus-edge
+              for i in $(kubectl -n $NAMESPACE get helmcharts.source.toolkit.fluxcd.io -o name); do kubectl -n $NAMESPACE patch $i --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'; done
```
amrc-benmorrow, line 23: The `edge-cluster` chart should definitely not be removing finalizers from Flux resources. Flux is not installed as part of this chart and will not be removed by deleting it. When we are talking about an edge cluster running on a cluster that already had Flux installed (like our on-prem soft-gateway clusters) Flux is not going to be removed at all.

We don't have a process for tearing down an edge cluster cleanly yet. There are multiple pieces that would need removing in a specific order; in particular removing the `edge-sync` deployment will leave any deployed Edge Agent charts behind. When necessary I tend to simply delete the namespaces and then clean up from there, but that most likely won't even run `helm uninstall` as the Flux `helm-controller` will have been terminated before it gets there.

However, making the `edge-cluster` chart uninstall cleanly, assuming Flux is still running to drive the uninstall process, would be a step in the right direction. All HelmRelease resources installed by `edge-sync` have the label `app.kubernetes.io/managed-by=edge-sync.factoryplus.app.amrc.co.uk`, so this post-delete hook could remove them too. This would tear down all the edge deployments, including the Edge Agents. Sealed secrets would be removed as part of the 'helm uninstall' if it was installed as a dep originally, and would be left alone otherwise; these are both correct.

Removing Flux (and deciding whether to remove it) is more complicated as it isn't installed via Helm in the first place, and Flux is driving this whole uninstall process. Flux itself and the Flux GitRepo/Kustomization link to the cluster-private repo were installed by the cluster bootstrap script, so properly we would want a cluster teardown script to handle removing it. I don't think this is in scope for now; if we can get to the point where deleting the `edge-cluster` Helm chart removes all the other F+ Helm charts and deployed resources I think that's a good step forward.

edge-helm-charts/charts/edge-cluster/templates/post-delete.yaml
```diff
@@ -0,0 +1,74 @@
...
+              for i in $(kubectl -n $NAMESPACE get GitRepository -o name); do kubectl -n $NAMESPACE patch $i --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'; done
+              for i in $(kubectl -n $NAMESPACE get HelmRelease -o name); do kubectl -n $NAMESPACE patch $i --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'; done
+              for i in $(kubectl -n $NAMESPACE get Kustomization -o name); do kubectl -n $NAMESPACE patch $i --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'; done
+              for i in $(kubectl -n $NAMESPACE get kerberos-keys -o name); do kubectl -n $NAMESPACE patch $i --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'; done
```
amrc-benmorrow, line 27: An Edge Agent deployed with external drivers will deploy LocalSecret resources. These are also managed by the krbkeys operator and need cleaning up the same as the KerberosKeys.

Currently no LocalSecret resources are ever deployed on a central cluster and the CRD is not deployed there, so trying to run `kubectl get localsecrets` will fail. (This will unexpectedly work on a cluster that has both central and edge clusters deployed, as CRDs are not namespaced.) Given that the central krbkeys is the same code as the edge and it may be useful to deploy LocalSecrets centrally later it is probably worth pulling in the CRD and ensuring they're deleted at uninstall now.

amrc-benmorrow (changes requested):

edge-helm-charts/charts/edge-cluster/templates/post-delete.yaml
djnewbould: Thanks for the information, I'll make some changes.

amrc-benmorrow (approved):

amrc-benmorrow:
I've moved the edge cluster conversations into a new ticket on our internal Jira.

Ideally this would be rebased onto `main` to remove the reverted commits, but I'm not going to hold up approving it for that.

PR #405, opened 2025-01-30
Fix key replacement logic in keytab trimming

djnewbould:
- Updated the calculation of key replacement times to ensure correctness.
- Added logic to track when each key version (KVNO) stops being the most recent, using the timestamp of the next higher KVNO.
- Defaulted keys with no replacements to "active" by assigning `float('inf')` as their replacement time.
- This ensures expired keys are removed based on accurate replacement timestamps rather than only their creation times.

This fix prevents premature removal of keys that may still be in use or valid until replaced by newer versions.

### DIFF
--- a/acs-krb-keys-operator/lib/amrc/factoryplus/krbkeys/keyops.py
+++ b/acs-krb-keys-operator/lib/amrc/factoryplus/krbkeys/keyops.py
@@ -84,13 +84,20 @@ class Keytab (KeyOps):
 
             principals = set(str(kte.principal) for kte in entries)
             for princ in principals:
-                mine    = [kte for kte in entries if str(kte.principal) == princ]
-                ckvno   = max((kte.kvno for kte in mine), default=-1)
-                atrisk  = [kte for kte in mine if kte.kvno < ckvno]
+                # Find keytab entries for current principle.
+                mine   = [kte for kte in entries if str(kte.principal) == princ]
+                # Sort entries by KVNO, then timestamp (oldest to newest).
+                mine.sort(key=lambda kte: (kte.kvno, kte.timestamp))
+                # Find the highest current version number for the principle.
+                ckvno  = max((kte.kvno for kte in mine), default=-1)
+                # Keys with a key version number less than ckvno.
+                atrisk = [kte for kte in mine if kte.kvno < ckvno]
+                # For each KVNO, find the creation timestamp of the next highest KVNO.
+                # If no higher KVNO exists, the key is "active"
                 replacement_times = {
                     kte.kvno: min(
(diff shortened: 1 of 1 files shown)


acs-krb-keys-operator/lib/amrc/factoryplus/krbkeys/keyops.py
```diff
@@ -84,10 +84,24 @@ def trim_keys (spec, secret):
...
-                mine    = [kte for kte in entries if str(kte.principal) == princ]
-                ckvno   = max((kte.kvno for kte in mine), default=-1)
-                atrisk  = [kte for kte in mine if kte.kvno < ckvno]
-                expired = [kte for kte in atrisk if kte.timestamp < since]
+                # Find keytab entries for current principle.
+                mine   = [kte for kte in entries if str(kte.principal) == princ]
+                # Sort entries by KVNO, then timestamp (oldest to newest).
+                mine.sort(key=lambda kte: (kte.kvno, kte.timestamp))
+                # Find the highest current version number for the principle.
+                ckvno  = max((kte.kvno for kte in mine), default=-1)
```
amrc-benmorrow, line 92: I think this line and line 94 are unnecessary now?

acs-krb-keys-operator/lib/amrc/factoryplus/krbkeys/keyops.py
```diff
@@ -84,10 +84,24 @@ def trim_keys (spec, secret):
...
+                # For each KVNO, find the creation timestamp of the next highest KVNO.
+                # If no higher KVNO exists, the key is "active"
+                replacement_times = {
+                    kte.kvno: min(
+                        (next_kte.timestamp for next_kte in mine if next_kte.kvno > kte.kvno),
+                        default=float('inf')
+                    )
+                    for kte in mine
+                }
+                expired = [kte for kte in atrisk if replacement_times[kte.kvno] < since]
```
amrc-benmorrow, line 104: ...you can just use `mine` here instead. The current KVNO will come out with an replacement time of `inf` and won't be considered for expiration.

amrc-benmorrow (changes requested):

acs-krb-keys-operator/lib/amrc/factoryplus/krbkeys/keyops.py
djnewbould: I left them because they're used when we check if they're any more old keys on line 111. I think we'd need to keep the logic to calculate that.

amrc-benmorrow: Duh. You're right.

amrc-benmorrow (approved):

PR #409, opened 2025-02-04
Backport ConfigDB version 2

amrc-benmorrow:
This pulls in just the ConfigDB changes from the auth rework branch. It would be good to get everyone working with the new ConfigDB if possible. This might mess up the merging but that will hopefully be fixable...


amrc-benmorrow:
Hang on, I'm going to hold off on this for a minute, it might be possible to merge the entire current state of the auth-rework branch. That would be better, and would avoid making a mess of the history.

PR #415, opened 2025-02-18
Migrate Directory Dumps to Service Setup

djnewbould:
**Please note: This branch will be merged into testing/v4**

Currently services register their service URLs individually with the Directory on startup. There are three mechanism for this:
- Some services publish an Sparkplug NBIRTH using the Service schema.
- Some services register themselves directly with the Directory HTTP API.
- Some services have an initContainer which uses curl to do the same thing.

This causes complication with each service needing to feed its own URL to register, and delay in registration which can cause other service to fail unnecessarily.

Service registration now happens when service setup runs. Service URLs are defined in the file `00_service-urls.yaml`, this directory dump file is loaded before the service client is initiated. The service client and loading of the other dump files are dependant on the directory services being loaded.

### DIFF
--- a/acs-directory/lib/api_v1.js
+++ b/acs-directory/lib/api_v1.js
@@ -102,9 +102,11 @@ export default class API {
     }
 
     async load_dump(req, res) {
-        console.log("called dump!");
-        await this.model.load_dump();
-        res.status(200).json({result: "It worked!"});
+        const dump = req.body;
+        const owner = req.params.owner
+           ?? await this.fplus.resolve_principal({kerberos: req.auth});
+        const status = await this.model.load_dump(dump, owner);
+        res.status(status).end();
     }
 
     async devices(req, res) {
--- a/acs-directory/lib/api_v1.js
+++ b/acs-directory/lib/api_v1.js
@@ -102,11 +102,13 @@ export default class API {
     }
 
     async load_dump(req, res) {
-        const dump = req.body;
-        const owner = req.params.owner
-           ?? await this.fplus.resolve_principal({kerberos: req.auth});
-        const status = await this.model.load_dump(dump, owner);
-        res.status(status).end();
+        try{
+            const info = await this.model.load_dump(req);
+            res.status(info).end();
+        }catch (e) {
+            console.log(e);
+            res.status(500).json({error: e.message});
+        }
     }
 
     async devices(req, res) {
--- a/acs-directory/lib/api_v1.js
+++ b/acs-directory/lib/api_v1.js
@@ -62,6 +62,8 @@ export default class API {
         api.get("/service", this.services.bind(this));
         api.get("/service/:service", this.service_providers.bind(this));
 
+        api.post("/load", this.load_dump.bind(this))
+
         api.route("/service/:service/advertisment")
             .get(this.service_advert_get.bind(this))
             .put(this.service_advert_put.bind(this));
@@ -97,6 +99,12 @@ export default class API {
     async search(req, res) {
         let list = await this.model.search(req.query);
         res.status(200).json(list);
+    }
+
+    async load_dump(req, res) {
+        console.log("called dump!");
+        await this.model.load_dump();
+        res.status(200).json({result: "It worked!"});
     }
 
     async devices(req, res) {
--- a/acs-directory/lib/api_v1.js
+++ b/acs-directory/lib/api_v1.js
@@ -102,6 +102,10 @@ export default class API {
     }
 
     async load_dump(req, res) {
+        if(req.auth !== this.fplus.opts.root_principal){
+            res.status(401).end();
+        }
+
         try{
             const info = await this.model.load_dump(req);
             res.status(info).end();
--- a/acs-directory/lib/api_v1.js
+++ b/acs-directory/lib/api_v1.js
@@ -65,14 +65,14 @@ export default class API {
         api.post("/load", this.load_dump.bind(this))
 
         api.route("/service/:service/advertisment")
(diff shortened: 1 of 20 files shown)


acs-service-setup/lib/uuids.js
```diff
@@ -69,6 +73,12 @@ export const ACS = {
      * changing to be per-deployment. */
     Device: {
         ConfigDB:               "36861e8d-9152-40c4-8f08-f51c2d7e3c25",
+        ClusterManager:         "127cde3c-773a-4f61-b0ba-7412a2695253",
```
amrc-benmorrow, line 76: Hmm, here we have a problem. This UUID is not fixed; it is generated by krbkeys as a per-installation UUID for this service. All the service accounts are meant to be moving to this scheme, it will be necessary if we ever get proper cross-realm support. For the specific case of the cluster manager, it has no Sparkplug interface, so I think the best thing to do for now is use the all-zeros UUID as the device. This can carry across to all the service registrations with no Sparkplug interface until the Directory is modified to make this optional.

Long-term the service Sparkplug interfaces need to go; the only other alternative would be to register HTTP interfaces via the dumps but Sparkplug interfaces via the MQTT ingester; this would be a bit messy but could be achieved.

amrc-benmorrow (changes requested):
This PR doesn't remove the service registration from the MQTT indexer. I think that is worth doing; it's always been a security hole.

acs-directory/lib/api_v1.js
```diff
@@ -62,6 +62,8 @@ export default class API {
         api.get("/service", this.services.bind(this));
         api.get("/service/:service", this.service_providers.bind(this));
 
+        api.post("/load", this.load_dump.bind(this))
+
         api.route("/service/:service/advertisment")
```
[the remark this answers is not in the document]
djnewbould: Just to double check, do you want me to remove the endpoints for ConfigDB and Auth as well as the directory?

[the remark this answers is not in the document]
amrc-benmorrow: Ah, no, I wasn't clear. This isn't about `/load`.
* Do move the Directory `load` to the top level.
* Don't touch the `load` endpoint's on Auth or ConfigDB, I'm dealing with those in WP1.
* This comment was about the `service/:service/advertisment` endpoints on the Directory, which should all be removed.
Basically, service registration should only be allowed via a dump.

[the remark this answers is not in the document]
djnewbould: Understood, Thank you 🙂

acs-directory/lib/loader.js
```diff
@@ -0,0 +1,27 @@
...
+export class Loader {
+    constructor (opts) {
+        this.model = opts.model;
+        this.fplus = opts.fplus;
+        this.routes = this.load.bind(this);
+    }
+
+    async load(req, res) {
+        if(req.auth !== this.fplus.opts.root_principal){
+            res.status(401).end();
```
amrc-benmorrow, line 16: This should be 403. 401 is for 'authentication failed', basically 'wrong password or expired token', and will prompt the client to reauthenticate. 403 is 'you've authenticated correctly but you're not allowed to do this'.

acs-directory/lib/model.js
```diff
@@ -130,23 +123,19 @@ export default class Model extends Queries {
         return true;
     }
 
-    async load_dump(req) {
+    async dump_load(req) {
         const dump = req.body;
         if (!this.dump_validate(dump)){
             return 400;
```
amrc-benmorrow, line 129: I think you copied this 400 from one of the other services; it's not correct. 400 is for malformed HTTP requests, including e.g. an `application/json` request with a body which isn't valid JSON. This situation is where the request was valid at the HTTP level but isn't in an appropriate format for this endpoint; this should return 422.

acs-directory/lib/model.js
```diff
@@ -130,23 +123,19 @@ export default class Model extends Queries {
...
         }
         for (const advertisement of dump.advertisements) {
-            const owner = advertisement.owner
-                ?? await this.fplus.resolve_principal({kerberos: req.auth});
-
-            if(!owner){
+            if(!advertisement.device){
```
amrc-benmorrow, line 132: If `device` is not optional then a missing device should fail validation rather than silently be ignored. But I think it's best to leave this for now; in my Auth branch I've set up a system of JSON schemas for validating the different dumps, with service-setup able to lint against those schemas, so it's probably better to migrate the Directory over to that system once it's merged to testing/v4.

acs-service-setup/dumps/service-urls.yaml
```diff
@@ -0,0 +1,27 @@
...
+    url: !url "http://directory.${namespace}.svc.cluster.local"
+    device: !u ACS.Device.Directory
+  - service: !u ACS.Service.ConfigDB
+    url: !url "http://configdb.${namespace}.svc.cluster.local"
+    device: !u ACS.Device.ConfigDB
+  - service: !u ACS.Service.CmdEsc
+    url: !url "http://cmdesc.${namespace}.svc.cluster.local"
+    device: !u ACS.Device.CmdEsc
+  - service: !u ACS.Service.Git
+    url: !url "http://git.${domain}"
```
amrc-benmorrow, line 17: This is not necessarily `http`. On installations with a certificate it must be `https`.

amrc-benmorrow (changes requested):
There is some reconcilation needed now with `testing/v4`; sorry about that. Try to rebase this branch onto `testing/v4` if you can.

amrc-benmorrow (approved):

PR #420, opened 2025-02-21
Add Internal Traefik Redirect

djnewbould:
The added coredns config map redirects all external URLs back to the internal Traefik service. This fixes issues caused by ACS instances which don't have external DNS setup.

### DIFF
--- a/deploy/values.yaml
+++ b/deploy/values.yaml
@@ -26,6 +26,10 @@ acs:
   # own section. Deployments from a Git checkout must set this value.
   #defaultTag: ''
 
+k3s:
+  traefikRedirect:
+    enabled: true
+
 identity:
   # -- Whether or not to enable the Identity component
   enabled: true
--- a/deploy/values.yaml
+++ b/deploy/values.yaml
@@ -26,7 +26,7 @@ acs:
   # own section. Deployments from a Git checkout must set this value.
   #defaultTag: ''
 
-coreDNS:
+coredns:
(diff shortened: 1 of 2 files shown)


deploy/values.yaml
```diff
@@ -26,6 +26,13 @@ acs:
   # own section. Deployments from a Git checkout must set this value.
   #defaultTag: ''
 
+k3s:
```
amrc-benmorrow, line 29: This is not k3s-specific. Although I don't think it's supported by vanilla K8s from Google it is supported by e.g. [Azure](https://learn.microsoft.com/en-us/azure/aks/coredns-custom), which is I think where k3s got the convention from. (The MS documentation is, as usual, vague on the specifics, so I don't know whether they support this `.override` feature or not.)

So I think this should be `coreDNS`.

deploy/values.yaml
```diff
@@ -26,6 +26,13 @@ acs:
...
+  # -- An option to enable redirecting all external URLs back to the
+  # internal Traefik service. ACS deployments without external DNS
+  # should enable this.
```
amrc-benmorrow, line 32: We need more explanation here. In particular this overrides a system resource (in the `kube-system` namespace); that needs noting clearly and unambiguously as people may have existing overrides in place for other reasons.

amrc-benmorrow (changes requested):

amrc-benmorrow (approved):

PR #432, opened 2025-03-06
Change Service Setup to Cron Job

djnewbould:
- Service setup now runs as a cron job that runs every 5 minutes rather than just at start-up and on helm upgrade. This ensures the cluster configuration maintains a desired state.
- Authorizers have been added to the hivemq plugin, these run on publish and subscribe to check if the principle has permission for the mqtt action. 

**This requires  more testing before merging**


hivemq-krb/src/main/java/uk/co/amrc/factoryplus/hivemq_auth_krb/ClientSessionStore.java
```diff
@@ -0,0 +1,21 @@
+package uk.co.amrc.factoryplus.hivemq_auth_krb;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ClientSessionStore {
+    private static final ConcurrentHashMap<String, String> sessionMap = new ConcurrentHashMap<>();
```
amrc-benmorrow, line 6: We don't need to use a global here. A ClientSessionStore instance can be created in FPKrbMain and passed down through the Providers. If necessary this class could be reworked to an FPKrbContext which holds both the ServiceClient and this session store map; this avoids passing down multiple context parameters.

hivemq-krb/src/main/java/uk/co/amrc/factoryplus/hivemq_auth_krb/FPKrbAuthorizer.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    public void authorizePublish(
+            @NotNull PublishAuthorizerInput publishAuthorizerInput,
+            @NotNull PublishAuthorizerOutput publishAuthorizerOutput
+    ) {
+        var clientId = publishAuthorizerInput.getClientInformation().getClientId();
+        var clientUsername = ClientSessionStore.getUsername(clientId);
+        var topic = publishAuthorizerInput.getPublishPacket().getTopic();
+        try{
```
amrc-benmorrow, line 39: You shouldn't need to try/catch here; errors will be asynchronous. You do want to use the second argument to `subscribe`, which is called for async errors.

hivemq-krb/src/main/java/uk/co/amrc/factoryplus/hivemq_auth_krb/FPKrbAuthorizer.java
```diff
@@ -0,0 +1,141 @@
...
+            isPermissionAllowed(getACLforPrincipal(clientUsername), topic, TopicPermission.MqttActivity.PUBLISH)
+                    .subscribe(result -> {
+                        if(result){
+                            publishAuthorizerOutput.authorizeSuccessfully();
```
amrc-benmorrow, line 43: Because you are handling this asynchronously you need to use the `async` method on the PublishAuthorizerOutput. Call `async` synchronously (not within Rx) to give you an Async, and then call the `getOutput` and `resume` methods on that object from within the Rx callback. Otherwise I think you end up blocking the whole broker...

hivemq-krb/src/main/java/uk/co/amrc/factoryplus/hivemq_auth_krb/FPKrbAuthorizer.java
```diff
@@ -0,0 +1,141 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.co.amrc.factoryplus.FPServiceClient;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class FPKrbAuthorizer implements SubscriptionAuthorizer, PublishAuthorizer {
```
amrc-benmorrow, line 22: A SubscriptionAuthorizer is only called when a client makes a new SUBSCRIBE request. If ACLs change while a client is subscribed these changes will not be picked up. I think the only way to handle this correctly is with the Interceptor API.

hivemq-krb/src/main/java/uk/co/amrc/factoryplus/hivemq_auth_krb/FPKrbMain.java
```diff
@@ -36,4 +40,26 @@ public void extensionStop(final @NotNull ExtensionStopInput extensionStopInput,
...
+                .registerServiceURL(FPUuid.Service.MQTT, url)
+                .retryWhen(errs -> errs
+                        .doOnNext(e -> {
+                            log.error("Service registration failed: {}", e.toString());
+                            log.info("Retrying registration in 5 seconds.");
+                        })
+                        .delay(5, TimeUnit.SECONDS))
+                .subscribe(() -> log.info("Registered service successfully"),
+                        e -> log.error("Failed to register service: {}",
+                                e.toString()));
```
amrc-benmorrow, line 61: I think this bit should have been removed? But we still need to call `http().start()`.

djnewbould: I've based this off of main instead of testing v4, it might be better to rebase to v4.

PR #438, opened 2025-03-13
Refactor Metric* imports to come from service-client

AlexGodbehere:
There were duplicates being exported from v2.0.0 of @amrc-factoryplus/utilities.

### DIFF
--- a/acs-cmdesc/package.json
+++ b/acs-cmdesc/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@amrc-factoryplus/utilities": "2.0.0",
+    "@amrc-factoryplus/service-client": "^1.5.0",
     "@isaacs/ttlcache": "^1.2.0",
     "express": "^4.18.1",
     "express-basic-auth": "^1.2.1",
--- a/acs-configdb/package.json
+++ b/acs-configdb/package.json
@@ -21,6 +21,7 @@
   "dependencies": {
     "@amrc-factoryplus/rx-util": "^0.0.3",
     "@amrc-factoryplus/service-api": "^0.0.2",
+    "@amrc-factoryplus/service-client": "^1.5.0",
     "@amrc-factoryplus/utilities": "2.0.0",
     "ajv": "^8.10.0",
     "ajv-formats": "^2.1.1",
--- a/acs-directory/lib/mqttcli.js
+++ b/acs-directory/lib/mqttcli.js
@@ -1,20 +1,16 @@
 /*
- * Factory+ / AMRC Connectivity Stack (ACS) Directory component service
- * MQTT client connection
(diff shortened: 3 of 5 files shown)


acs-cmdesc/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@amrc-factoryplus/utilities": "2.0.0",
+    "@amrc-factoryplus/service-client": "^1.5.0",
```
AlexGodbehere, line 26: Oh wait, we don't need this if v2.0.0 exports *?

acs-configdb/package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "@amrc-factoryplus/rx-util": "^0.0.3",
     "@amrc-factoryplus/service-api": "^0.0.2",
+    "@amrc-factoryplus/service-client": "^1.5.0",
```
AlexGodbehere, line 24: Or this.

acs-cmdesc/package.json
amrc-benmorrow, line 26: I would pull in 1.3.6 here. I know it's not the latest but `utilities` pulls in 1.3.4 and we don't want duplicate copies in `node_modules`.

acs-directory/lib/mqttcli.js
```diff
@@ -1,20 +1,16 @@
...
-    SpB, Topic, UUIDs
-} from "@amrc-factoryplus/utilities";
-import { Device_Info } from "./constants.js";
-import { Schema } from "./uuids.js";
+import timers from 'timers/promises'
+import async from 'async'
+import Long from 'long'
+
+import { Address, SpB, Topic, UUIDs } from '@amrc-factoryplus/utilities'
+import { MetricBranch, MetricBuilder } from '@amrc-factoryplus/service-client'
```
amrc-benmorrow, line 10: Is MetricTree missed on purpose? I though the Directory used that...

amrc-benmorrow (changes requested):

acs-cmdesc/package.json
AlexGodbehere: Do you mean to state `1.3.6` and `1.3.4` here? Won't that result in two versions?

acs-directory/lib/mqttcli.js
AlexGodbehere: Good spot.

acs-cmdesc/package.json
amrc-benmorrow: Sometimes Node is funny about explicitly importing a dep of a dep. It does funny things with nested `node_modules` dirs. IME it's better to explicitly depend on every module you `import`.

amrc-benmorrow (approved):

PR #442, opened 2025-03-14
Implement Auth Editor on the Admin UI

Kallb123:
Brings the Auth editor into the new Admin UI.

image

- View lists of: Principals, Groups, and Permissions.
- For each Principal: manage the groups they are members of, manage the permissions they are granted, and see the effective permissions they are granted.
- For each Group: manage the members of the group and manage the permissions the groups is granted.
- For each Permission: manage the entries within the Auth service which use the permission and see what groups the permission is a member of.
- Link a Kerberos principal to a F+ User UUID


amrc-benmorrow (approved):
I've pushed a small change to add a display column for the plural field on permission grants.

PR #443, opened 2025-03-17
Deploy Keycloak as an OAuth provider

### LINKED ISSUE #273: Proposal: Add OAuth2 Support to Auth Service
Our current auth service relies on Kerberos principals and ACLs to authenticate and authorise users but we are increasingly encountering applications and services that support only OAuth2 for authorisation (e.g. Grafana). To improve compatibility between these applications and ACS, I propose adding OAuth2 support.

**Benefits:**
- **Enhanced Compatibility:** Many modern applications and third-party services support OAuth2. Adding OAuth2 support will enable these applications to integrate seamlessly with ACS.

**Proposed Implementation:**

1. **OAuth2 Authorisation Server:**
   - Implement an OAuth2 authorisation server within our existing auth service.
   - Support the following OAuth2 grant types:
     - Authorisation Code Grant
     - Implicit Grant
     - Resource Owner Password Credentials Grant
     - Client Credentials Grant
   - Ensure the authorisation server can issue access tokens, refresh tokens, and handle token revocation.

2. **Token Handling:**
   - Implement token generation, validation, and expiration mechanisms.
   - Store tokens securely, possibly using encryption and hashing techniques to enhance security.
   - Ensure compatibility with JWT (JSON Web Tokens) for token structure and payload.

3. **Integration with Existing Authentication:**
   - Integrate OAuth2 flows with our existing Kerberos-based authentication mechanism.
   - Use Kerberos principals to authenticate users during the OAuth2 authorisation process.
   - Map Kerberos ACLs to OAuth2 scopes to ensure appropriate authorisation levels.

4. **Documentation and Examples:**
   - Provide comprehensive documentation on how to integrate an OAuth2 application with ACS.
   - Include example code and configurations for common scenarios and applications.

**Resources:**
- [OAuth 2.0 RFC](https://tools.ietf.org/html/rfc6749)
- [Node-OAuth2-Server](https://oauth2-server.readthedocs.io/en/latest/)

KavanPrice:
This adds an OAuth mechanism to ACS with Kerberos integration.
- Defines "OpenID" deployment and service
- Defines IngressRoute for the service
- Creates a ConfigMap for a basic startup realm
  - Realm name: "factory_plus"
  - Realm clients:
    - grafana-oauth
  - Realm user federation:
    - Kerberos
- Adds the `auth.generic_oauth` section to `grafana.ini`

Only the client for the Grafana service is implemented here.


deploy/templates/openid/realm-config.yaml
```diff
@@ -0,0 +1,2740 @@
...
+        ],
+        "org.keycloak.storage.UserStorageProvider": [
+          {
+            "id": "237e7955-1144-4130-a9a0-ec324e1725ab",
+            "name": "Kerberos",
+            "providerId": "kerberos",
+            "subComponents": {},
+            "config": {
+              "serverPrincipal": [
+                "HTTP/openid.{{.Values.acs.baseUrl}}@{{.Values.acs.baseUrl | upper}}"
```
amrc-benmorrow, line 1988: `.Values.identity.realm`. Realms don't necessarily follow the standard convention; in particular our v2 installation didn't.

amrc-benmorrow (changes requested):

KavanPrice:
- Removed secret lookup and helm templating from Values
- Add service setup step for
  - Blank realm creation with Kerberos support
  - Dynamic client creation for clients defined in Values
- Remove `realm-config` monolith

acs-service-setup/lib/openid-realm.js
```diff
@@ -0,0 +1,247 @@
...
+   * Run setup for the realm. This generates the full realm representation.
+   *
+   * A basic realm is created first and then populated with clients.
+   *
+   * @async
+   * @returns {Promise<void>}
+   */
+  async run() {
+    let base_realm = {
+      id: crypto.randomUUID(),
```
amrc-benmorrow, line 42: What happens the second time this runs? Will it try to create a new realm?

acs-service-setup/lib/openid-realm.js
```diff
@@ -0,0 +1,247 @@
...
+      components: {
+        "org.keycloak.storage.UserStorageProvider": [
+          {
+            id: crypto.randomUUID(),
+            name: "Kerberos",
+            providerId: "kerberos",
+            subComponents: {},
+            config: {
+              serverPrincipal: [
+                `HTTP/openid.${this.base_url}@${this.base_url.toUpperCase()}`,
```
amrc-benmorrow, line 56: We have `acs_config.realm` for this. The convention of using the uppercased domain is only a convention; our on-prem V2 installation didn't follow it.

acs-service-setup/lib/openid-realm.js
```diff
@@ -0,0 +1,247 @@
...
+        throw new Error(`${status}: ${error}`);
+      }
+    } catch (error) {
+      this.log(
+        `Couldn't get an initial access token for realm setup: ${error}`,
+      );
+    }
+  }
+
+  async wait(milliseconds) {
```
amrc-benmorrow, line 244: `timers/promises` ?

deploy/templates/openid/openid.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+apiVersion: factoryplus.app.amrc.co.uk/v1
+kind: KerberosKey
+metadata:
+  name: sv1openid
+  namespace: {{ .Release.Namespace }}
+spec:
+  type: Random
+  principal: sv1openid@{{ .Values.identity.realm | required "values.identity.realm is required!" }}
+  secret: openid-keytabs/client
```
amrc-benmorrow, line 91: Do we need this? Is Keycloak a client of any F+ services?

deploy/templates/openid/local-secrets.yaml
```diff
@@ -0,0 +1,11 @@
+{{if .Values.openid.enabled}}
+apiVersion: factoryplus.app.amrc.co.uk/v1
+kind: LocalSecret
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: "keycloak-grafana-client"
+spec:
+  format: Password
+  secret: "keycloak-grafana-client"
+  key: "grafana"
```
amrc-benmorrow, line 10: This needs to generate multiple keys, one per client we are registering. For now I think we keep them all in the same Secret; this isn't ideal but makes the deployment much easier. So
* We need a `{{ range $k, $v := .Values.serviceSetup.config.openidClients }}` which generates multiple LocalSecret resources referencing different keys in the same Secret.
* The secret wants renaming to something more generic like `keycloak-clients`.
* The `extraSecretMounts` under the Grafana values needs an `items` property restricting Grafana to only read the `grafana` secret.

You could reasonably remove the `redirectHost` property and just use the key for the client service hostname.

acs-service-setup/lib/openid-realm.js
```diff
@@ -0,0 +1,247 @@
...
+   * @param {Boolean} is_retry - Whether the request is a retry after a 401. If this is false and a 401 is returned, this will retry after refreshing the token.
+   * @returns {Promise<void>} Resolves when the client is created.
+   */
+  async create_client(client_representation, is_retry) {
+    const secret_path = path.join(
+      "/etc/secret",
+      client_representation.redirectHost,
+    );
+    const content = await fs.readFile(secret_path, "utf8");
+    const client_secret = content.trim();
```
amrc-benmorrow, line 139: Is there any way to configure Keycloak to read the secret from its own filesystem at runtime, rather than reading the file contents here? This would be more reliable against the secret being updated. (In principle we should rotate all our secrets regularly.)

amrc-benmorrow (changes requested):

acs-service-setup/lib/openid-realm.js
amrc-benmorrow: A second container on the keycloak pod that called into the API to update the secrets every hour would also be a reasonable solution.

acs-service-setup/lib/openid-realm.js
KavanPrice: Ah, I'm bad at JS - I didn't know this existed.

KavanPrice: Fixed in [9a80e55](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/pull/443/commits/9a80e558ed5885c071a1cc8675657c3be182c421)

deploy/templates/openid/openid.yaml
KavanPrice: Not yet. Removed in [15ed6a7](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/pull/443/commits/15ed6a7419c46d520b9b72c8f1a528c895fd747b)

acs-service-setup/lib/openid-realm.js
amrc-benmorrow: OK. So we can't update anything, but that's OK for MVP. As long as service-setup finishes correctly in that case I think that's OK.

deploy/templates/openid/local-secrets.yaml
KavanPrice: Done in [e935fbb](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/pull/443/commits/e935fbbca50a83d3ef34ebb757fd8b618a4450b8)

KavanPrice: There was a bug here with accessing the namespace. Fixed in [e0b480f](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/pull/443/commits/e0b480f6e13657b2a21dece9a3c5a05f310c2b36)

acs-service-setup/lib/openid-realm.js
KavanPrice: I'm looking at this. There are apparently ways of making it read a secret from a vault. Simplest solution might be, as you suggested, a container that just updates the secrets

AlexGodbehere:
Superseded by https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/pull/646

PR #445, opened 2025-03-17
Use local libs for acs-admin

amrc-benmorrow:
Update acs-admin to use the in-tree `service-client`. This should also allow other in-tree packages like `rx-client` to be used.


Kallb123:
Can we get this merged into `feature/admin-ui` as that is like the base feature branch which the auth and manager are on. I guess we can cherry pick across if needed? Not entirely sure the best workflow :)

amrc-benmorrow:
Like that?

amrc-benmorrow:
For workflow, I would either:
* Merge the current work for Auth and Manager before this PR, then pick up new branches afterwards, OR
* Merge this PR and then rebase the current Auth and Manager work onto the new `feature/admin-ui`.

But this is yours and Alex's branch so you may prefer to just merge `feature/admin-ui` into your individual branches and carry on.

Kallb123 (approved):
Works on my Windows machine after enabling Developer Mode in order to allow symlinks to be made as a standard user. We may want to report to bun that the non-symlink method appears to not be working.

PR #446, opened 2025-03-17
service-setup for files service - draft

amrc-za:


acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,46 @@
...
+---
+service: !u UUIDs.Service.ConfigDB
+version: 2
+objects:
+  !u Auth.Class.ServiceRole:
+    !u Files.Requirement.ServiceAccount:
+      name: "Files Service server"
+      subclassOf:
+        - !u Auth.Class.CentralService
+        - !u ACS.Group.SparkplugNode
```
amrc-benmorrow, line 11: I don't think the Files service has a Sparkplug interface? I don't think it should have, if it needs to expose notify it should use the new WebSocket interface, but with all files immutable there's nothing to change.

acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,46 @@
...
+    !u Files.Service.Files: { name: "Files hosting service" }
+  !u UUIDs.Class.App:
+    !u Files.App.Config: { name: "Files Configuration" }
+  !u Files.Perm.All:
+    !u Files.Perm.Upload: { name: "Upload" }
+    !u Files.Perm.Download: { name: "Download" }
+
+  !u ConfigDB.Class.R1Class:
+    !u Files.Class.File: { name: "File" }
+    !u Files.Class.FileType: { name: "File Type" }
```
amrc-benmorrow, line 29: _File type_ is a rank 2 class containing subclasses of _File_. This means it needs creating under `ConfigDB.Class.R2Class`.

acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  !u Files.Class.FileType:
+    !u Files.FileType.PDF: { name: "PDF" }
```
amrc-benmorrow, line 32: Each of these needs marking as a subclass of _File_.

acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,46 @@
...
+    !u Files.FileType.CSV: { name: "CSV" }
+    !u Files.FileType.CAD: { name: "CAD" }
+
+---
+service: !u UUIDs.Service.Authentication
+version: 2
+grants:
+  !u Files.Requirement.ServiceAccount:
+    !u Files.Perm.All:
+      !u Files.Service.Files: false
```
amrc-benmorrow, line 42: This should be a grant to allow the Files service to read its own ACLs. This means it should be
```
!u Files.Requirement.ServiceAccount:
  !u Auth.Perm.ReadACL:
    !u Files.Perm.All: true
```
That is: the files service has permission to read ACLs using any of the permissions in the _Files permission_ class.

acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  !u ACS.Group.Administrator:
+    !u Files.Perm.All:
+      !u Files.Service.Files: false
```
amrc-benmorrow, line 46: This should be
```
!u ACS.Group.Administrator:
  !u Files.Perm.All: null
```
which is a wildcard grant of all the files service permissions. What you had was granting permission to do Files service operations on the Files service registration UUID, which isn't a _File_ or a _File type_ and so wouldn't have done anything useful.

amrc-benmorrow (changes requested):
A few changes needed here.

Now we have a _Service role_ for the Files service the KerberosKey in `deploy/templates/files/files.yaml` needs updating to put the `sv1files` account in that group. See e.g. `deploy/templates/cluster-manager/krbkeys.yaml` for an example. The `class` of the `sv1files` account should be _Central service_, and it should be in the group which is the service role you just created.

acs-service-setup/dumps/files.yaml
amrc-za: updated

acs-service-setup/dumps/files.yaml
amrc-za: updated

acs-service-setup/dumps/files.yaml
amrc-za: - made the `Files.Class.FileType` itself the subclass of File (`Files.Class.File`)
- made each `Files.FileType`: PDF, CSV, CAD a subclass of File (`Files.Class.File`)

acs-service-setup/dumps/files.yaml
amrc-za: - created the `Files.Class.FileType` under the `ConfigDB.Class.R2Class`
- made the `Files.Class.FileType` a subclass of `Files.Class.File`

acs-service-setup/dumps/files.yaml
amrc-za: removed the `ACS.Group.SparkplugNode`

amrc-za:
`sv1files` account added to files.yaml within deploy/templates

acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,57 @@
...
+    !u Files.Perm.Download: { name: "Download" }
+
+  !u ConfigDB.Class.R1Class:
+    !u Files.Class.File: { name: "File" }
+
+  !u ConfigDB.Class.R2Class:
+    !u Files.Class.FileType:
+      name: "File Type"
+      subclassOf:
+        - !u Files.Class.File
```
amrc-benmorrow, line 33: This subclass relation is not correct and will not load because the ranks are incorrect. I'm not sure if you've seen the new ConfigDB object model docs? https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/blob/testing/v4/docs/object-model.md

_File_ is a rank 1 class, a class of individuals. _File type_ is a rank 2 class, a class of classes of individuals, so _File type_ cannot be a subclass of _File_. The members of _File type_, e.g. _PDF_, are subclasses of _File_, but _File type_ itself is not.

Strictly speaking we should have a third class here, _File group_, which is the powerset of _File_; that is, the class of all subclasses of _File_, or the class of all possible groups of files. Then _File type_ is a subclass of _File group_: it's specifically those groups of files which are grouped by file type. If we wanted e.g. 'groups of files where each group is associated with a particular device' then that would be a different subclass of _File group_. This matches the existing pattern of _Principal_ / _Principal group_ / _Principal type_ and _Permission_ / _Permission group_ / _Service permission set_.

amrc-benmorrow (changes requested):

acs-service-setup/dumps/files.yaml
amrc-za: I made changes:
`File` - Rank 1
Actual file types such as: `PDF, CAD, CSV, TXT` - Rank 1 and subclasses of `File`

`FileGroup` - Rank 2
`FileType` - Rank 2 subclass of `FileGroup`

acs-service-setup/dumps/files.yaml
```diff
@@ -0,0 +1,62 @@
...
+  !u ConfigDB.Class.R2Class:
+    !u Files.Class.FileGroup: {name: "File Group"}
+    !u Files.Class.FileType:
+      name: "File Type"
+      subclassOf:
+        - !u Files.Class.FileGroup
+
+  !u ConfigDB.Class.R1Class:
+    !u Files.Class.File: { name: "File" }
+    !u Files.FileType.PDF:
```
amrc-benmorrow, line 37: These should all still be members of _File type_, that is, they should be like
```
!u Files.Class.FileType:
  !u Files.FileType.PDF:
```
rather than sitting as direct members of _Rank 1 class_. This still makes them rank 1 as _File type_ is rank 2 and a class member is always one rank lower than the class. They will also still be members of _Rank 1 class_: as _File type_ is rank 2 and has no explicit superclasses it will be automatically made a subclass of _Rank 1 class_.

I realise this is confusing; from a traditional OO background this way of looking at membership and subclassing seems rather odd.

amrc-za: fixed

PR #449, opened 2025-03-18
Br/file service

BryanRam:


acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,150 @@
...
+      await this.configDb.put_config(App.Config, file_uuid, fileJSON);
+      console.log('Put to config completed.');
+      return res.status(201).json({ message: 'File uploaded successfully' });
+    } catch (err) {
+      console.error('Error when posting file: ', err.message);
+      res.status(500).end();
+    }
+  };
+
+  isFileTypeSupported(file_type_uuid) {
```
amrc-benmorrow, line 142: We certainly don't want a fixed list of file types. But see above; classification into types should happen outside the Files service anyway.

amrc-benmorrow (changes requested):

acs-files/bin/api.js
```diff
@@ -0,0 +1,43 @@
...
+import { multer_storage } from '../config/multer.js';
+
+const { env } = process;
+
+const fplus = await new ServiceClient({
+  env,
+}).init();
+
+const configDb = await new ConfigDB(fplus);
+const auth = await new Auth(fplus);
```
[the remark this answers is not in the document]
amrc-za: fixed

acs-files/bin/api.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+import { ServiceClient } from '@amrc-factoryplus/service-client';
+import { WebAPI } from '@amrc-factoryplus/service-api';
+import { ConfigDB, Auth } from '@amrc-factoryplus/service-client/interface.js';
```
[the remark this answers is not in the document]
amrc-za: removed intefaces

acs-files/bin/api.js
```diff
@@ -0,0 +1,43 @@
...
+const configDb = await new ConfigDB(fplus);
+const auth = await new Auth(fplus);
+
+multer({ storage: multer_storage });
+
+const api = await new WebAPI({
+  ping: {
+    version: Version,
+    service: Service.Registry,
+    device: env.DEVICE_UUID,
```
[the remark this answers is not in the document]
amrc-za: fixed

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,150 @@
...
+      );
+
+      if (!ok) return res.status(403).end();
+
+      const fileObj = await this.configDb.get_config(App.Config, file_uuid);
+
+      // File not found in ConfigDb
+      if (!fileObj || !fileObj.file_uuid) {
+        return res.status(404).json({ message: 'File not found in ConfigDb' });
+      }
```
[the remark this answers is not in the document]
amrc-benmorrow: We don't have `DELETE /v1/file/:uuid` at this point... let's call this out of scope for now, unless it's trivial to implement.

acs-files/config/multer.js
```diff
@@ -0,0 +1,18 @@
+import multer from 'multer';
+import { v4 as uuidv4 } from 'uuid';
+import * as dotenv from "dotenv";
+dotenv.config({});
```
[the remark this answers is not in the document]
amrc-za: replaced dotenv with dotenvx

acs-files/bin/api.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+import { ServiceClient } from '@amrc-factoryplus/service-client';
+import { WebAPI } from '@amrc-factoryplus/service-api';
+import { ConfigDB, Auth } from '@amrc-factoryplus/service-client/interface.js';
+import { routes } from '../lib/routes.js';
+import multer from 'multer';
+import { multer_storage } from '../config/multer.js';
```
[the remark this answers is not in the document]
amrc-za: multer removed, accepting files from plain req.body

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,150 @@
...
+
+      return res
+        .status(200)
+        .download(process.env.FILES_STORAGE + '/' + file_uuid, (err) => {
```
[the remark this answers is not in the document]
amrc-za: Removed direct access to process.env from files that are not bin/api.js. 
They are passed as opts.

acs-files/lib/api-v1.js
amrc-za: Completely removed

acs-files/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Use an official Node.js runtime as a parent image
+FROM node:18
```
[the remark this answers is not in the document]
amrc-za: Dockerfile done,
Makefile not done yet

acs-files/bin/api.js
```diff
@@ -0,0 +1,43 @@
...
+      revision: GIT_VERSION,
+    },
+  },
+  debug: fplus.debug,
+  realm: env.REALM,
+  hostname: env.HOSTNAME,
+  keytab: env.SERVER_KEYTAB,
+  http_port: env.PORT,
+  max_age: env.CACHE_MAX_AGE,
+  body_limit: env.BODY_LIMIT,
```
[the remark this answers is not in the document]
amrc-za: Added `app.use(express.raw({ type: '*/*', limit: this.body_limit })); ` to WebAPI

acs-files/lib/api-v1.js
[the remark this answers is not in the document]
amrc-benmorrow: No, that's not the correct permission check, we need a new permission Perm.ListStorage. This should be checked with `check_acl(req.auth, Perm.ListStorage, Special.Null, false)` and should always be granted with the wildcard UUID. (Basically, it's a permission that doesn't have a target, so we always use the wildcard.)

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,141 @@
...
+      return res
+        .status(400)
+        .json({ message: 'FAILED: File Uuid not provided.' });
+
+    if (!Valid.uuid.test(file_uuid))
+      return res.status(410).json({ message: 'FAILED: File Uuid is invalid' });
+
+    // Check auth permission
+    // Todo: disable wildcard in the future
+    // Todo: update Target to be individual File, at the moment it checks if Principal has Download permission to the whole FileType object under which all Files are stored.
```
amrc-benmorrow, line 59: * Wildcard should **not** be disabled. We want to accept wildcard grants in the Auth service.
* The target should be `file_uuid`.

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,141 @@
...
+
+    if (!ok)
+      return res
+        .status(403)
+        .json({ message: 'FAILED: No Download permission' });
+
+    const filePath = path.join(this.uploadPath, file_uuid);
+
+    // Check if file exists
+    fs.access(filePath, fs.constants.F_OK, (err) => {
```
amrc-benmorrow, line 75: Don't do this, it introduces a race condition. If the file is created or deleted between this `access` call and the `open` call made by the actual file-sending function then we will have sent the wrong answer.

You should be using `sendFile` rather than `download` in any case; `download` is more browser-compatibility stuff. The documentation is not clear but I believe `sendFile` will correctly send a 404 for you when the file doesn't exist. You need to test this.

For now set the Content-Type of the response to `application/octet-stream`; properly we should be recording the Content-Type of the request which created the file and returning it with the response, but that would mean recording that information somewhere, either in the ConfigDB or locally in our own storage; in either case we're back with the problem of the ConfigDB and the Files storage getting out of sync. There are solutions here but we've taken so long to get to this point we can't really explore them.

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,141 @@
...
+
+  async postFile(req, res) {
+    // Todo: disable wildcard in the future.
+    // Check ACL exists for the Principal to Upload to FileType. This means that the uploaded file objects will be stored under FileType (Rank2) object.
+
+    const ok = await this.auth.check_acl(
+      req.auth,
+      Perm.Upload,
+      App.Config,
+      true
```
amrc-benmorrow, line 97: This is now another permission with no target, so is checked with `(req.auth, Perm.Upload, Special.Null, false)`. This check is simply for permission to create a file at all; we don't have any other information to make a permission check on.

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,141 @@
...
+        message: 'FAILED: No Upload permission',
+      });
+    }
+
+    // Create object in ConfigDB of Class File
+    const created_uuid = await this.configDb.create_object(Class.File);
+
+    // Upload file to Storage
+
+    fs.mkdirSync(this.uploadPath, { recursive: true });
```
amrc-benmorrow, line 111: Don't ever use the `Sync` calls in a web backend, they'll block the whole process. I would use `fs.promises` and await the result, always. You can use the callback API but that makes the code more confusing.

acs-files/lib/api-v1.js
```diff
@@ -0,0 +1,141 @@
...
+
+    fs.writeFile(file_path, req.body, async (err) => {
+      if (err) {
+        res.status(500).json({ message: 'FAILED: File upload failed.' });
+      }
+
+      const stats = fs.statSync(file_path); // to get file size
+
+      const currDate = new Date();
+      const original_file_name = req.headers['x-filename'] || null;
```
amrc-benmorrow, line 122: Where does this come from?

lib/js-service-api/lib/webapi.js
```diff
@@ -29,63 +29,65 @@ export class WebAPI {
...
+      let app = this.app;
+
+      /* CORS */
+      app.use(cors({ credentials: true, maxAge: 86400 }));
+
+      /* Body decoders. These will only decode request bodies of the
+       * appropriate content-type. */
+      app.use(express.json({ limit: this.body_limit, strict: false }));
+      app.use(express.text({ limit: this.body_limit }));
+      app.use(express.raw({ type: '*/*', limit: this.body_limit })); // Parses raw binary bodies
```
amrc-benmorrow, line 41: We need options here. Most of the services want JSON decoding on JSON content-types; the Files service wants raw decoding on everything (including JSON content-types). So there needs to be an option to WebAPI, e.g. `body_type: "json"` or `body_type: "raw"` which defaults to `"json"` for compatibility. Then we switch on that option here and install the appropriate decoders.

(The `text` decoder can be removed altogether, it's not used. It's there from a very long time ago...)

lib/js-service-api/lib/webapi.js
```diff
@@ -29,63 +29,65 @@ export class WebAPI {
     }
 
     async init () {
-        let app = this.app;
-
-        /* CORS */
-        app.use(cors({ credentials: true, maxAge: 86400 }));
+      let app = this.app;
```
amrc-benmorrow, line 32: Please don't change the indentation. There are more code style changes below; please stop your editor from doing this on other people's code.

acs-files/lib/api-v1.js
amrc-za: the original name should be passed in headers, that's the only way to get it. If it's not manually passed the originalName will be null.

acs-files/lib/api-v1.js
amrc-za: fixed

acs-files/lib/api-v1.js
amrc-benmorrow: OK. You're correct, of course, I'm just not sure an `X-` header is the best answer here. It doesn't matter; leave it as it is for now.

acs-files/lib/api-v1.js
BryanRam: Removed check for if file exists and replaced `res.download` with `res.sendFile`. Also set the Content-Type of the response to `application/octet-stream`. While testing what responses I get when downloading a file, it does return a 404 when the file doesn't exist so can confirm that part.

acs-files/lib/api-v1.js
amrc-za: changed to be 'original-filename'

lib/js-service-api/lib/webapi.js
amrc-za: - removed text decoded
- added options
- defaulting to JSON

PR #458, opened 2025-03-24
Restrict edge-sync to spec patching

### LINKED ISSUE #457: Helm Release Not Uninstalled When HelmRelease CR is Deleted
### Summary

When an `EdgeDeployment` is deleted, the associated `HelmRelease` CR is removed as expected. However, the actual Helm release (and its pods) is **not** uninstalled by Flux.

This is caused by a conflict between `edge-sync` and Flux: `edge-sync` removes finalisers from the CR, while Flux attempts to add them back. Depending on the exact timing of deletion, Flux may never see the removal and therefore not uninstall the release.

### Behaviour

- ❌ HelmRelease CR is deleted  
- ❌ Helm release and pods are **not** uninstalled  
- 🛑 Flux never sees the finaliser and does not reconcile the deletion  
- ⚠️ This is **not version-dependent** — the difference between working and broken behaviour was due to timing, not changes in code.

### Root Cause

`edge-sync` removes the `finalizers` array from the `HelmRelease` CR. If the CR is deleted while the finaliser is absent, Flux does not detect the deletion and cannot perform cleanup.

This leads to orphaned Helm releases, with running pods and resources remaining in the cluster even though the CR is gone.

### Steps to Reproduce

1. Deploy a Node/Edge Agent and register with the Manager.
2. Create an `EdgeDeployment`.
3. Delete the `EdgeDeployment` before Flux has a chance to re-add the finaliser.

### Notes

- The issue is race-condition based: it depends on whether Flux has re-added the finaliser before the CR is deleted.
- Confirmed in multiple versions — this is a general issue in the way `edge-sync` interacts with Flux.

### Proposed Solution

Update `edge-sync` to **preserve Flux’s finalisers** on HelmRelease CRs. It should not remove or override them unless explicitly

AlexGodbehere:
This prevents edge-sync from removing Flux's finalizers.

This was causing a bug where sometimes a HelmRelease resouce would be deleted but Flux would not uninstall the actual helm release; this was caused by the missing finalizer.

Fixes #457.

### DIFF
--- a/acs-edge-sync/lib/deployments.js
+++ b/acs-edge-sync/lib/deployments.js
@@ -47,15 +47,19 @@ class Reconciliation {
                 if (!deep_equal(old.spec, man.spec)) {
                     // Create a patch that only updates the spec
                     const patch = {
-                        spec: man.spec
+                        apiVersion: "helm.toolkit.fluxcd.io/v2beta1",
+                        kind: "HelmRelease",
+                        spec: man.spec,
+                        metadata: {
+                            name,
+                            namespace: this.deploy.namespace
+                        }
                     };
                     this.log("PATCH: %o", patch);
                     await this.check(this.objs.patch(
-                        name,
-                        this.resource.apiVersion,
-                        this.resource.kind,
-                        this.deploy.namespace,
                         patch,
+                        undefined,
+                        undefined,
                         undefined,
                         undefined,
                         { headers: { 'Content-Type': 'application/merge-patch+json' } }
--- a/acs-edge-sync/lib/deployments.js
+++ b/acs-edge-sync/lib/deployments.js
@@ -1,7 +1,5 @@
 /*
- * Factory+ / AMRC Connectivity Stack (ACS) Edge Sync Op
- * Deployment operator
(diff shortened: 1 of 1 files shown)


djnewbould (commented):
After I delete a device, I get this error on edge-sync. 
```
/home/node/app/node_modules/@kubernetes/client-node/dist/object.js:364
            throw new Error('Required spec property kind is not set');
                  ^

Error: Required spec property kind is not set
    at KubernetesObjectApi.specUriPath (/home/node/app/node_modules/@kubernetes/client-node/dist/object.js:364:19)
    at KubernetesObjectApi.patch (/home/node/app/node_modules/@kubernetes/client-node/dist/object.js:167:41)
    at Reconciliation.apply (file:///home/node/app/lib/deployments.js:53:48)
    at Deployments.reconcile_manifests (file:///home/node/app/lib/deployments.js:230:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.11.1
```

amrc-benmorrow (approved):

PR #466, opened 2025-03-26
ConfigDB Editor within the Admin UI

Kallb123:
Bringing the ConfigDB editor interface into the new Admin web portal.

## Config Editor

The config entry editor seems to work fine in both JSON and YAML, however, there is no error highlighting or auto-complete. It does show a bit of syntax colours. I think there are problems with web workers and bundlers and things that I don't understand! It doesn't throw errors, so it's "fine" but it's not reached its full potential.

There was also a statement within `monaco-yaml` which said it only works with bundlers. I worry that this might break builds, but I haven't been able to try it. You could remove d7970cf9ea8dc227be110b75e1bf6b46efb06f4e if it causes problems

## TODO

- [x] Most of the user interface for: Application list (searchable) -> Config entry list -> Config editor
- [ ] Tidy up the header of these pages (use Alex's Manager breadcrumb UI)
- [x] Switching the config entry list to a searchable data table
- [ ] Create application user interface (can create an object anyway)
- [x] Create application config entry interface
- [x] Implement the Monaco Editor to edit the config entries, including buffer logic for incoming changes ("do you want to accept remote" prompt if input is dirty).Being able to press save to push the config entry.
- [x] Toggling between JSON+YAML. `pinia-persist` is used to save user preference.
- [x] Monaco YAML support
- [ ] Monaco error highlighting (need help with web workers I think, would be nice to test in an https environment)
- [x] Methods within the config entry list to delete an entry
- [x] Most of the user interface for: Object list (searchable) -> Clicking an object to see classifications and member relationships
- [x] Add object interface on the object list (select a class, enter name and uuid)
- [x] Object selector with only single select
- [x] Methods within object page to add/remove memberships and members
- [x] Fix bug where direct memberships don't update
- [x] Methods within object page to add/remove classifications
- [x] Method within object page to allow setting the primary class
- [ ] Add object of this class interface on the object page (can create an object manually, this is shortcut)
- [ ] Fix bug within data tables where columns can't be toggled
- [x] Change directRelationship store to use RX
- [ ] JSON Dumps interface (out of scope)


amrc-benmorrow (approved):
I've pushed a few changes based on testing.

PR #469, opened 2025-03-27
Fix Manager Route Guard

djnewbould:
- [x] Add login page 
- [x] Add Auth Route guard.
- [x] Fix logout pino state reset.
- [x] Fix login failure logic.


AlexGodbehere:
Thanks @djnewbould!

AlexGodbehere (approved):

PR #474, opened 2025-04-09
Add support for processing array payloads in device metrics

AlexGodbehere:
# Add support for JSON array payloads in MQTT

## Description
This PR adds support for handling JSON array payloads in MQTT messages. Previously, when receiving an array of objects (e.g., `[{value: 25, timestamp: 1234}, {value: 26, timestamp: 1235}]`), only the first object would be processed. With this change, all objects in the array are processed as separate metric updates.

## Changes
- Added new utility functions in `typeHandler.ts`:
  - `isArrayPayload`: Detects if a payload is an array of objects
  - `processArrayPayload`: Processes each object in an array and extracts values and timestamps
- Modified the `_handleData` method in `device.ts` to process all objects in array payloads
- Simplified `parseValueFromPayload` and `parseTimeStampFromPayload` to focus on non-array payloads
- Added support for objects with 'value', 'timestamp', or 'time' properties

## Example
With this change, if a metric receives a JSON array payload like:
```json
[
  {"value": 25.5, "timestamp": 1625097600000},
  {"value": 26.0, "timestamp": 1625097601000},
  {"value": 26.5, "timestamp": 1625097602000}
]
```

then multiple metrics will be emmited from the edge agent - one per sample.

Google Chrome 2025-04-10 08 49 18


amrc-benmorrow (changes requested):
I can see this is a useful feature, but it cannot be implemented like this.

* This will corrupt non-JSON payloads which happen to look like valid JSON, for instance a Buffer which happens to contain `0x31313131` (`1111`).
* Guessing what to do with the data based on whether we happen to be returned an array or not is very fragile and will cause trouble later.
* The `value` property for the objects in the array is unconfigurable, just like `timestamp`.

I think this might be worth attacking in three stages.

Initially we just need to get it working (I presume you have an immediate need for this). But rather than messing up the whole decoding process bu guessing we declare what we're doing: create a new payload format, `JSON array`. This accepts only and exactly a JSON array of objects with `value` and `timestamp` properties.

Next we can re-introduce some flexibility, based on these properties of JSONPath:
* A JSONPath expression always starts with `$` and never contains `;`.
* JSONPath allows extracting an array of results using `[*]`.

So we extend the meaning of the JSON `path` property
* The basic form of a JSON `path` propery is changed to `readings=R;value=V;timestamp=T`. The `R`, `V`, `T` are JSONPath expressions. This is parsed like a query-string as name-path pairs.
* A path expression with an initial `$` is treated as `value=PATH` for compatibility.
* The expression `R` is evaluated to return an array of readings. Then `V` and `T` are evaluated against each reading to return the value and timestamp.
* If `readings` is missing we use the whole value as a single reading.
* If `timestamp` is missing we use our idea of the current time.
* Timestamps must be sanity-checked.

The `JSON array` format then goes back to being JSON format with a path of `readings=$[*];value=$.value;timestamp=$.timestamp`. We can consider adding more properties later, like maybe `timestampFormat`.

As a final step, we can and should generalise this by replacing `path` with `arrayPath`, `valuePath` and `timestampPath`. This would allow useful things like reading multiple values with timestamps out of a Buffer. Unless we can just generalise this query-string format for the path property across all payload types? I'm not sure whether there would be any conflicts with existing valid paths.

PR #478, opened 2025-04-11
Fixes and active testing

AlexGodbehere:
Not complete yet, but the findings so far are in DRI-298. Once merged we can cut a beta/RC.


acs-admin/src/store/serviceClientStore.js
```diff
@@ -33,7 +33,7 @@ export const useServiceClientStore = defineStore('service-client', {
         this.loaded = true;
         this.scheme  = import.meta.env.SCHEME;
         this.baseUrl = import.meta.env.BASEURL;
-
+        client.Fetch.cache = 'reload';
```
Kallb123, line 36: This was just removed #468 
Are we sure we want to disable caching?

AlexGodbehere: I was having issues with stale data, but I can't remember exactly where so not very helpful, sorry!

AlexGodbehere, line 36:
```suggestion
        // client.Fetch.cache = 'reload';
```

acs-admin/vite.config.js
```diff
@@ -36,15 +38,16 @@ export default defineConfig({
     }
   },
   plugins: [
+    importMetaEnv.vite({
+      env: '.env',
+      example: '.env.example',
+      files: ['dist/index.html'], // explicitly define the correct path
+    }),
```
AlexGodbehere, line 45: @Kallb123 what does this fix?

acs-admin/src/store/serviceClientStore.js
```diff
@@ -33,7 +33,7 @@ export const useServiceClientStore = defineStore('service-client', {
         this.loaded = true;
         this.scheme  = import.meta.env.SCHEME;
         this.baseUrl = import.meta.env.BASEURL;
-        client.Fetch.cache = 'reload';
+        // client.Fetch.cache = 'reload';
```
AlexGodbehere, line 36: I've enabled caching again.

acs-admin/.gitignore
```diff
@@ -15,6 +15,8 @@ coverage
 *.local
 .env
 
+bun.lock
```
AlexGodbehere, line 18: Why?

acs-admin/vite.config.js
Kallb123: After fixing the `json-schema-ref-parser` error showing `exports not defined` on build, there was another error that showed up: `TypeError: Object prototype may only be an Object or null: import_meta_env_placeholder`.
This indicates that the `import-meta-env` package isn't behaving as intended. Apparently this can happen when it is loaded after the `vite` plugin, so bringing it earlier in the plugin list is meant to help it be more consistent. It could be that it differs on different platforms etc, so being explicit makes it more consistent.

Kallb123: I think this was again only something within a prod build. I don't think the `import-meta-env` stuff is used during `bun dev`

PR #483, opened 2025-04-16
Add draft spec for file service octant

djnewbould:
The added spec defines a solution for the automatic ingestion of TDMS files into ACS.


amrc-benmorrow:
There are a few problems here, mostly to do with not thinking through which machines are running which processes. It probably be a good idea to draw a deployment diagram showing machines, clusters and containers and showing how the data flows from one place to another. Be particularly careful to distinguish files created on the host machine from files created within a container.

## Component 1: Folder watcher

This is not (cannot be) part of the File Service itself; it cannot run on the central cluster. If we are interacting with a directory on which TDMS files are being generated by some external process then at minimum we are on an edge cluster, and possibly not on a Factory+ machine at all. So questions of using `child_process` and JS vs Python are not relevant.

This is a stand-alone script of some kind; it needs to run on whichever machine is actually generating the TDMS files. The script will be a client of the File Service or of Influx as appropriate; also of the ConfigDB. There are some non-trivial questions here around authentication, particularly if this runs off-cluster, but they can be solved.

If the TDMS files are generated on a Cell Gateway machine then this can be deployed via the V3 Edge Deployment mechanism, though this will probably mean mapping a host path into the container which is not entirely trivial. Otherwise it will need to run external to the F+ infrastructure and will need a username/password for interacting with F+. If it wants to access Influx directly then we don't have a solution for that yet so we will need to work out how to create it an appropriate token.

Monitoring the status of this component is also non-trivial as it will almost certainly be impossible to directly connect to the machine it's running on. I would suggest it should report its status over Sparkplug; this can easily include raising an Alert if there is a problem. This could either be implemented directly or the Folder Watcher component could communicate with an Edge Agent running on the cluster. As an alternative it could post updates to the ConfigDB.

## Component 2: Uploader

Option 1 here would be implemented within the Folder Watcher component: that component would simply upload the file to the File Service. Option 2 would be more awkward to implement from the edge; it might be better to start with Option 1 and then look into possibilities for post-processing the data centrally after it's been uploaded. The ConfigDB change-notify interface can be used to watch for new TDMS files and then a central service can download the file and perform processing on it.

## Component 3: Summary generator

This job can be also be performed by a central post-processing service. The summary record can be generated once and pushed to the ConfigDB.

## Component 4: Storage

I would not consider introducing more storage at this point, and certainly not storage at the edge; this just creates problems. The File Service provides blob storage and the central Influx instance provides timeseries storage. If it isn't convenient to use these services for a job like this then we need to provide the services so that it is.

amrc-benmorrow:
A third option to consider, depending on the volume of data here, would be to split this up further:

* Run an MQTT broker on a Cell Gateway (via a V3 deployment).
* Run a component off-cluster which publishes the TDMS files to MQTT, along with its own status.
* Run a component on-cluster which reads the TDMS files from MQTT and pushes them to the File Service. This can also report its status via MQTT.
* Run an Edge Agent on-cluster which publishes the status information over Sparkplug.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+The POST file endpoint will parse a header containing the UUID of the file extension. The change/notify interface will be implemented on 
+the API; on file creation it will emit the file UUID and the file type UUID. 
+
+### File Edge Agent
+The File Edge Agent will be created to automatically upload files TDMS files from the edge. It will be a NodeJS application
+which watches a directory for new files, when a file is created in this directory, the file will be posted to the ACS File Service. 
+This application will run off cluster and use the JS-Service-Client with username and password authentication. 
+
+### Summary Generator
+The Summary Generator is a Python application which will run in the ACS-Files pod on the central cluster. The application will
```
amrc-benmorrow, line 27: This does not need to run in the Files pod; this is a separate service. It can run entirely separately. Potentially this doesn't need to be part of ACS at all, it could be deployed onto the cluster separately.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+listen to the change/notify interface of the ACS File Service API, when a new file is created and the file type is TDMS, the
```
amrc-benmorrow, line 28: Hmm, there is a problem here: at the point where the File Service creates the _Files Configuration_ entry the file object hasn't yet been classified as TDMS. However, the summariser can instead watch for members of the TDMS file type class and ensure they all get processed. This also avoids unnecessary work checking on non-TDMS uploads.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+![Deployment Diagram](deployment.png)
+
+### File Upload process
+
+1. The TDMS file is created by the data producer and stored in a directory on the local file system.  
+2. The File Edge Agent "folder watcher" detects a new file has been created and queues the file to be posted to the 
+ACS file service.  
+3. The queue is processed and the TDMS file is posted by the File Edge Agent "uploader" to the ACS File Service.  
+4. The ACS File Service API creates the file metadata in ConfigDB.  
+5. The ACS File Service API Writes the uploaded TDMS file to the Kubernetes volume.
```
amrc-benmorrow, line 43: 5a. Once the File Service POST request has returned a file object UUID, the uploader uses the ConfigDB API to classify this object as a TDMS file.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+
+### File Upload process
+
+1. The TDMS file is created by the data producer and stored in a directory on the local file system.  
+2. The File Edge Agent "folder watcher" detects a new file has been created and queues the file to be posted to the 
+ACS file service.  
+3. The queue is processed and the TDMS file is posted by the File Edge Agent "uploader" to the ACS File Service.  
+4. The ACS File Service API creates the file metadata in ConfigDB.  
+5. The ACS File Service API Writes the uploaded TDMS file to the Kubernetes volume.   
+6. The change/notify interface on the ACS File Service sends the created file UUID and the file type UUID.
```
amrc-benmorrow, line 44: 6. The change-notify interface on the **ConfigDB** notifies the Summary Generator that a new TDMS file has been created.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+7. On the file created notification, the File Service Summary Generator looks up the type UUID in ConfigDB to check if the file 
+is a TDMS file.  
+8. If the type is TDMS, a get request is made to the File Service API to retrieve the file.  
+9. The summary is generated by the File Service Summary Generator and inserted into InfluxDB.  
+10. The summary is visible in Grafana using the Influx datasource.   
+
+## Future Work
+The above defines an architecture for handling the storage of high frequency data files in ACS but has the following drawbacks:
+- The File Edge agent is not part of the edge cluster and can't be managed by the central cluster.
+- The summary won't be visible in near real-time due to the summary being generated after the file has been uploaded.
```
amrc-benmorrow, line 54: The delay here will be minimal. As soon as the ConfigDB classification has been made the summariser will get to work.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+# File Service Octant
+
+## Objective
+
+- Define process for the automatic ingestion of large TDMS files from the local file system directory into
+the ACS File Service storage.
+
+- Investigate if ACS can extract a data summary from the TDMS files.
+
+- The application must support the uploading of file being produced at a rate ranging from 1-10+ minutes and sizes ranging from 100MB to 14GB.
```
amrc-benmorrow, line 11: If we are looking at files this big then the File Service implementation needs changing. Currently the entire POST body will be read into a Buffer, with a size limit; instead the code needs to stream to the file, ideally with no size limit, or possible something large just for sanity checking. This means disabling the body decoders altogether and using the Node streams interface on the Request object.

The uploading client will need to work the same way. Also currently the Files service volume is deployed as 20G.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+The above defines an architecture for handling the storage of high frequency data files in ACS but has the following drawbacks:
+- The File Edge agent is not part of the edge cluster and can't be managed by the central cluster.
+- The summary won't be visible in near real-time due to the summary being generated after the file has been uploaded.
+- The Summary generator will have to be modified to support the parsing of new data file types.
+
+The proposed solution:
+- Create a new edge agent type: **File Edge Agent** which will be a Sparkplug Node. 
+- The summary generator will be a driver, these can be created for the parsing of different file types and will follow a 
+similar architecture to the existing Edge Agent. 
+- Instead of inserting the data summary into influx directly, device summary data will be published to MQTT.
```
amrc-benmorrow, line 61: If it's possible and acceptible to do this, i.e the summary data alone is sufficient and is a small enough quantity to be published over Sparkplug, then this would be much simpler to implement altogether. The whole watch/summarise process can run as a driver for the existing Edge Agent. It is possible to run Edge Agent drivers off-cluster, though setting up the authentication isn't fully explored yet.

Potentially the off-cluster summariser could both be an Edge Agent driver publishing summaries and also upload complete files to the File Service. This might raise some questions about whether drivers should be able to authenticate to the Edge Agent using Kerberos credentials; currently they can't, but in principle it could be supported.

amrc-benmorrow (changes requested):

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+- Investigate if ACS can extract a data summary from the TDMS files.
+
+- The application must support the uploading of file being produced at a rate ranging from 1-10+ minutes and sizes ranging from 100MB to 14GB.
+The files will contain data from the sensors written at a high frequency (1MHz).
+ 
+## Solution 
+
+### ACS File Service
+The ACS File service deployed in ACS v4.0.0 will be modified to store file extensions in the files metadata stored in ConfigDB.
+The POST file endpoint will parse a header containing the UUID of the file extension. The change/notify interface will be implemented on
```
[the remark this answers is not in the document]
BryanRam: The original idea here was to store the TDMS file extension UUID as part of the metadata for the file, so we would have a file_type field in the json object that gets registered in ConfigDB, so we could know that we should classify that file as a TDMS file type. Basically adding to the json object like this:

     let file_JSON = {
      file_uuid: created_uuid,      
      date_uploaded: curr_date,      
      user_who_uploaded: req.auth,      
      file_size: stats.size,      
      original_file_name: original_file_name,      
      file_type: file_type_uuid, //add file type as part of metadata      
      };

If instead we can just monitor (using **change-notify?**) when a new entry is added to ConfigDB and then classify it into the TDMS file type, then this change would not be necessary.

docs/concepts/File Service/file-service-octant.md
```diff
@@ -0,0 +1,62 @@
...
+
+- The application must support the uploading of file being produced at a rate ranging from 1-10+ minutes and sizes ranging from 100MB to 14GB.
+The files will contain data from the sensors written at a high frequency (1MHz).
+ 
+## Solution 
+
+### ACS File Service
+The ACS File service deployed in ACS v4.0.0 will be modified to store file extensions in the files metadata stored in ConfigDB.
+The POST file endpoint will parse a header containing the UUID of the file extension. The change/notify interface will be implemented on 
+the API; on file creation it will emit the file UUID and the file type UUID.
```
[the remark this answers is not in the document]
BryanRam: @amrc-benmorrow 
So would the intended workflow here for the client be:
- Client uploads file via Files Service
- Files Service creates corresponding entry in ConfigDB 
(new steps below)
- Client watches ConfigDB for any changes in Files Configuration
- Client gets new file UUID once a change has happened 
- Client creates child entry using file UUID in TDMS file type?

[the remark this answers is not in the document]
amrc-benmorrow: I am assuming 'client' here is the Uploader component and not anything running on the central cluster. The exact intended workflow is:

* The Uploader makes a POST request to the Files service containing the content of the TDMS file.
* The Files service creates a new ConfigDB _File_ object.
* The Files service creates a _Files Configuration_ metadata entry for the new _File_.
* The Files service commits the file data to the K8s volume.
* The Files service returns a response to the POST containing the UUID of the new File object.
* The Uploader has been awaiting this response and now knows the UUID of the file it just created.
* The Uploader knows this was a TDMS file and uses the ConfigDB API to add the new _File_ to the _TDMS_ class.

docs/concepts/File Service/file-service-octant.md
djnewbould: We're just double checking the size of the summary to see if we can publish over MQTT but I think this is the route we're going to go down. On driver authentication, is the flow similar to this?:
- Driver connects to Edge Agent MQTT server using username and password auth.
- Edge Agent check credentials with central auth service. 
- If  successful, authorise drive to publish to the Edge agent.

djnewbould: Also, is the new driver something we want in the main ACS repo or would it be better placed in a new repo of the FactoryPlus organisation?

amrc-benmorrow: Currently the authentication is strictly cluster-local for simplicity and security. So:
* The edge-agent Helm chart deploys a LocalSecret resource for each driver.
* This prompts the edge krbkeys to generate a Secret.
* The edge-agent and driver are both on-cluster and have access to this Secret.

This will need to change to properly support off-cluster drivers. Probably the flow you describe is the correct one, but we need to think whether to keep the existing system for on-cluster drivers or make them use Kerberos identities as well, and there will need to be changes to the Edge Agent to support authentication against Kerberos. For get-it-working hacks I can talk you through how to get hold of credentials which will work.

If we are publishing this publically I think it goes in ACS. Almost every time we have put something outside of ACS we've ended up pulling it in later. It can live on a long-term branch for a while until we're happy it's releasable.

PR #487, opened 2025-04-23
Add router guard to Admin UI

Kallb123:
A router guard was semi-implemented, but disabled due to it malfunctioning.

The logic is now corrected and the user should be directed to the correct routes every time.

This still stores the user credentials in localStorage, this will be needed until other authentication mechanisms are in place.

### DIFF
--- a/acs-admin/src/main.js
+++ b/acs-admin/src/main.js
@@ -43,7 +43,8 @@ const routes = [
     component: Login,
     meta: {
       name: "Login",
-      icon: "user-circle"
+      icon: "user-circle",
+      anonAllowed: true,
     }
   },
   {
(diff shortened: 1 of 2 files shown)


acs-admin/src/main.js
```diff
@@ -131,21 +132,49 @@ const router = createRouter({
...
+    return
+  }
+  // From here on, we do need auth
+  // If we are navigating between pages with the service client ready
+  if (useServiceClientStore().ready) {
+    next()
+    return
+  }
+  // Otherwise we are loading fresh and need to initialise the service client
+  const optsString = localStorage.getItem('opts');
```
amrc-benmorrow, line 151: Given this localStorage item is written from the serviceClientStore, I would have put all the logic from here to the end into a `useServiceClientStore().tryLogin()` method. This would return a boolean which tells us whether we go `next()` or `next({path: "/login"})`. Then all localStorage access is in the same place. But this isn't important right now.

amrc-benmorrow (approved):

acs-admin/src/main.js
Kallb123: That does make sense and would be tidy. I'll consider that in future :)

PR #493, opened 2025-04-28
Refactor useNodeStore to make use of useStore() for automatic updates

AlexGodbehere:


AlexGodbehere:
This was the last store (for the manager functionality, anyway) still using the old store architecture.

PR #508, opened 2025-05-12
Improve UI of ConfigDB Component

AlexGodbehere:
A few small UI tweaks to the ConfigDB component. I was aiming to add breadcrumbs like we have in the edge manager pages but didn't get around to it.

image
image


AlexGodbehere:
@Kallb123 ignore the screenshots saying `Config Store`, we've revered to `ConfigDB`.

Kallb123 (approved):
Looks much tidier this way, nice one!

PR #520, opened 2025-05-30
Za/files service tdms service setup

amrc-za:
- Added TDMS file type to the Files Service service setup.
- Added node_modules/ and .env to the main .gitignore


amrc-benmorrow:
This is not going to be merged to ACS.

PR #521, opened 2025-05-30
Add File Service Interface

djnewbould:
Adds a service interface in the js-service-client for the file service. The interface adds methods to upload and download files from the file service from both node.js and browser. 

- [x] Create Service Interface
- [x] Add directory service registration for the file service.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -214,6 +214,7 @@ jobs:
           - edge-modbus
           - edge-test
           - edge-tplink-smartplug
+          - edge-ads
     permissions:
       contents: read
       packages: write
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ config.mk
 */.npm_install_done
 .vscode
 .DS_Store
+/edge-ads/node_modules
--- a/acs-edge/.gitignore
+++ b/acs-edge/.gitignore
@@ -34,3 +34,4 @@ build/
 local_secrets
 
 edge_agent_config.yaml
+driver-passwords
--- a/acs-service-setup/dumps/edge.yaml
+++ b/acs-service-setup/dumps/edge.yaml
@@ -51,6 +51,7 @@ objects:
     !u ACS.Driver.Test:             { name: "Test" }
     !u ACS.Driver.TPlinkSmartPlug:  { name: "TP-Link smart plug" }
     !u ACS.Driver.External:         { name: "External driver" }
+    !u ACS.Driver.ADS:              { name: "ADS" }
 
 configs:
   # Driver Definition Configuration
@@ -562,6 +563,76 @@ configs:
         required:
           - host
         title: TP-Link Smart Plug Details
+        type: object
+    !u ACS.Driver.ADS:
+      image:
+        repository: edge-ads
+      polled: false
+      presentation:
+        address:
+          description: "The TwinCAT symbol name with optional cycle time (format:
+            symbol_name[,cycle_time_ms])"
+          icon: microchip
+          placeholder: e.g. Global_IO.iAnalogue,500
+          title: Symbol Name
+        path:
+          description: Must be set to $
+          icon: code
+          placeholder: Must be set to $
+          title: Path
+      schema:
+        properties:
+          localAdsPort:
+            description: Local ADS port (auto-assigned if not specified)
+            options:
+              order: 6
+            title: Local ADS Port (Optional)
+            type: number
+          localAmsNetId:
+            description: Local AMS NetId (auto-assigned if not specified)
+            options:
+              order: 5
+            title: Local AMS NetId (Optional)
+            type: string
+          routerAddress:
+            description: IP address of the ADS router (usually the PLC)
+            minLength: 1
+            options:
+              order: 2
+            title: Router Address
+            type: string
+          routerTcpPort:
+            default: 48898
+            description: TCP port of the ADS router
+            options:
+              order: 4
+            title: Router TCP Port
+            type: number
+          targetAdsPort:
+            default: 851
+            description: ADS port of the target (851 for TwinCAT runtime)
+            options:
+              order: 3
+            title: Target ADS Port
+            type: number
+          targetAmsNetId:
+            description: AMS NetId of the target PLC (e.g. 172.16.48.37.1.1)
+            minLength: 1
+            options:
+              order: 1
+            title: Target AMS NetId
+            type: string
+          timeoutDelay:
+            default: 5000
+            description: Connection timeout in milliseconds
+            options:
+              order: 7
+            title: Timeout Delay (ms)
+            type: number
+        required:
+          - targetAmsNetId
+          - routerAddress
+        title: ADS Connection Details
         type: object
 
     !u ACS.Driver.External:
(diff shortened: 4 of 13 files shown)


lib/js-service-client/lib/service-client.js
```diff
@@ -140,4 +140,5 @@ ServiceClient.define_interfaces(
     ["Fetch", SI.Fetch, `fetch`],
     ["Git", SI.Git, ``],
     ["MQTT", SI.MQTTInterface, `mqtt_client`],
+    ["Files", SI.Files, `files`],
```
amrc-benmorrow, line 143: The third entry `files` here tries to make `client.Files.files` also available as `client.files` directly. This was for compatibility with very old version of the library that didn't have the service interface objects. Don't  use this for new interfaces, just pass the empty string.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+        this.service = Service.Files;
+        this.file_types = new Map([
+            ["pdf",  Class.PDF],
+            ["txt",  Class.TXT],
+            ["csv",  Class.CSV],
+            ["CAD",  Class.CAD],
+        ])
+    }
+
+    async fetch (opts) {
```
amrc-benmorrow, line 20: Can these changes here not be integrated into the method on ServiceInterface?

I made a poor design choice using `body` for a JON body; for the Python interface I used `body` for a raw body or `json` for a JSON body with implied content-type. But that can't be changed now; perhaps a `raw` option, or perhaps use `type-is` to check for `application/json` or `+json` and JSON-encode if so. (We must support `+json` types, ConfigDB PATCH uses `application/merge-patch+json`.)

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+        }
+        if (st != 200){
+            this.throw("Can't list files", st);
+        }
+        return await res.json();
+    }
+
+    /**
+     * Downloads a file linked to the passed UUID and writes it to disk.
+     * @platform Node.js only - Not supported in browsers
```
amrc-benmorrow, line 61: It would be useful to have a method which returns a stream of the file contents which can be used on browsers. This would not be a Node stream, but a web streams API stream; Node supports both these days.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+            accept: 'application/octet-stream',
+        });
+        const st = res.status;
+        if (st == 404){
+            return;
+        }
+        if (st != 200){
+            this.throw(`Error finding file: ${uuid}`, st);
+        }
+        const outputPath = path.resolve(output_path, fileName);
```
amrc-benmorrow, line 85: I would not do this, I would just accept the full filename from the caller. Or allow either a string which is used directly or a function which is called with the original name and the uuid and returns a filename?

Given the method has no return value it could usefully return the original file name.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+     * Uploads a file to the file service using the passed file object uuid.
+     * @platform Node.js only - Not supported in browsers
+     * @param uuid The UUID of the file object.
+     * @param filePath Path of the file to upload.
+     */
+    async upload_file_with_uuid(uuid, filePath){
+        const isNode = typeof window === 'undefined' && typeof process !== 'undefined';
+        if(!isNode){
+            this.throw("Method not supported in browser");
+        }
```
amrc-benmorrow, line 101: Use `this.fplus.opts.browser` here. Detection like this is unreliable; if we decide on a reliable detection mechanism it can be used to set the `browser` option automatically.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+        this.throw(`Error uploading file: ${uuid}`);
+    }
+
+    /**
+     * Uploads a file to the file service from the browser.
+     * @platform Browser only - Not supported in Node.js
+     * @param file The file to upload.
+     * @param uuid UUID of the file object.
+     */
+    async upload_file_with_uuid_from_browser(file, uuid){
```
amrc-benmorrow, line 129: Why are these parameters the other way round?

Can this not be integrated into the method above? We can do `instanceof` checks on the `file` parameter to see what it is.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+            this.throw("Method not supported in browser");
+        }
+        const uuid = await this.fplus.ConfigDB.create_object(Class.File);
+        if(!uuid){
+            this.throw(`Error creating file object`);
+        }
+        await this.upload_file_with_uuid(uuid, filePath);
+        const fileName = path.basename(filePath);
+        const extension = fileName.split(".")[1];
+        const fileTypeClass = this.file_types.get(extension.toLowerCase());
```
amrc-benmorrow, line 192: Never depend on file extensions for anything. Get the caller to give you a file type UUID.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+/** Interface to the File service. */
+export class Files extends ServiceInterface {
+    constructor(fplus) {
+        super(fplus);
+        this.service = Service.Files;
+        this.file_types = new Map([
+            ["pdf",  Class.PDF],
+            ["txt",  Class.TXT],
+            ["csv",  Class.CSV],
+            ["CAD",  Class.CAD],
```
amrc-benmorrow, line 16: This sort of hardcoded list is a red flag. If this information was needed then it should be pulled from the ConfigDB so it can be extended. But it's not; the caller should be supplying an appropriate classification.

For the current use case the file type used probably won't just be 'TDMS files', it will be 'TDMS files from this device'. This can be made a subclass of _TDMS file_ as part of the device deployment. That's why it's important the uploader accepts a custom file type in its configuration.

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
...
+        if(st == 201) return;
+        this.throw(`Error uploading file: ${uuid}`);
+    }
+
+    /**
+     * Creates a file object, uploads the file to the file service and configures the file type.
+     * @platform Browser only - Not supported in Node.js
+     * @param file File to upload.
+     */
+    async upload_file_from_browser(file){
```
amrc-benmorrow, line 156: As above; surely these two methods can be integrated?

lib/js-service-client/lib/service/files.js
```diff
@@ -0,0 +1,196 @@
+import {ServiceInterface} from "./service-interface.js";
+import {App, Class, Service} from "../uuids.js";
+import fs from 'fs';
+import * as path from "node:path";
+import stream from "node:stream";
```
amrc-benmorrow, line 5: Have you verified this still webpacks correctly with these Node-specific imports? This probably means test builds of both the visualiser and the admin interface; they both use the service-client library in the brower and they pack it differently. You may need to change this to use dynamic imports instead.

Alternatively you could just accept/return [Web Streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) and leave it to the caller to handle interaction with the filesystem. Node has conversions to/from Node streams.

amrc-benmorrow (changes requested):

lib/js-service-client/lib/service/files.js
djnewbould: Both builds seem to work fine. I've built them with the makefile and they both work without issue.

djnewbould: I ended up adding dynamic imports because I was concerned other bundlers might have issues, so this way it should defiantly work with all of them 🙂

amrc-benmorrow (approved):

PR #523, opened 2025-06-03
Br/edge files tdms summariser

BryanRam:
Creates a TDMS Summariser class that calls a python child process to parse a TDMS file, then return JSON data as a summary.


amrc-za (changes requested):
TDMSsummariser:
(1)
- `filePath `should not be passed into constructor.
- Create an event handler method `handleFile({filePath})`  - here make sure the parameter is `object `within `{}`
- Move your summary generation logic from `run()` into the `handleFile` 
- Inside the `run()` method, subscribe to FILE_UPLOADED event (see how it's done in my `Uploader `class). 
(2)
- Test it by creating an instance of `TDMSSummariser` in the `main.js` within main method by calling its `run()` function before `uploader.run()`

PR #529, opened 2025-06-10
Add JSON (Batched) payload format for batch metric processing

AlexGodbehere:
Implements a new "JSON (Batched)" payload format to safely handle batch JSON metrics without the fragile auto-detection that was corrupting non-JSON payloads.

## Problem

The existing implementation attempted to auto-detect JSON arrays within the regular JSON payload format, which:
- Corrupted non-JSON payloads that happened to look like valid JSON (e.g., Buffer containing "1111")
- Used fragile guessing logic based on payload structure
- Made the value property unconfigurable for array objects

## Solution

Added a new explicit "JSON (Batched)" payload format that:
- Requires users to explicitly select the format when sending batch data
- Accepts only exact JSON array format: `[{value: X, timestamp: Y}, ...]`, taking path into account (`$.value`). Timestamp is currently hard coded.
- Processes batches at the device level rather than in individual value parsing
- Eliminates all auto-detection and guessing logic

### Core Implementation
- **typeHandler.ts**: Added `JSONBatched` enum value and `processJSONBatchedPayload()` function
- **device.ts**: Removed auto-detection logic, added explicit JSONBatched handling in `_handleData()`
- **REST.ts**: Updated to handle JSONBatched format with correct Content-Type

### Configuration & UI
- **schema.json**: Added "JSON (Batched)" to payload format enum
- **DeviceConnectionForm.vue**: Added "JSON (Batched)" option to UI dropdown

### Cleanup
- Removed deprecated `processArrayPayload()` and `isArrayPayload()` functions
- Removed fragile auto-detection code that was causing payload corruption

## Usage

Users can now:
1. Select "JSON (Batched)" as payload format in Manager UI
2. Send payloads like: `[{"value": 23.5, "timestamp": 1640995200000}, {"value": 24.1, "timestamp": 1640995260000}]`
3. Timestamps are required (auto-generated if missing, however this is problematic for batched payloads and should be provided).
4. Timestamps in seconds are automatically converted to milliseconds

## Breaking Changes

None - this is purely additive. Existing "JSON" format behavior is unchanged.

## Future Work

This implementation provides the foundation for:
- Stage 2: Enhanced JSONPath support with query-string format
- Stage 3: Generalized path support across all payload types


acs-edge/lib/helpers/typeHandler.ts
```diff
@@ -472,6 +562,10 @@ export function parseTimeStampFromPayload(msg: any, metric: sparkplugMetric, pay
                 log('Failed to parse JSON data')
             }
 
+            // Note: Array payloads are now handled in _handleData using processArrayPayload
+            // This function only processes non-array payloads
+
+            // Extract timestamp from the payload
```
amrc-benmorrow, line 568: I would seriously consider removing this feature altogether and just insisting that southbound sources that want to set their own timestamps use the batched format.

amrc-benmorrow (approved):

PR #531, opened 2025-06-16
Allow publishing JS libs to NPM

amrc-benmorrow:
Change the GH actions to support publishing the in-tree JS libs to NPM. Support this using release tags of the form `js/lib-name/vN.N.N` and a new top-level workflow to choose which form of release this is. The existing release workflow no longer runs on `release` but on `workflow_call`.


djnewbould (dismissed):
Looks good! I didn't know you call a workflow from another workflow.

PR #543, opened 2025-07-08
Edge Files TDMS Driver Code Review

amrc-za:


edge-files-tdms/lib/tdms-file-summariser.js
```diff
@@ -0,0 +1,84 @@
...
+    this.bindToEvents();
+  }
+
+  bindToEvents() {
+    this.eventManager.on(EVENTS.FILE_UPLOADED, this.handleFileSummary.bind(this));
+  }
+
+  async handleFileSummary({filePath}) {
+    try {
+     if (!(await isFileExist(filePath))) {
```
amrc-benmorrow, line 22: This is not useful. If the file doesn't exist an error will be thrown by TdmsFile in the Python. More importantly, there's a race condition: just because the file exists right now doesn't mean it still will when Python tries to open it.

edge-files-tdms/lib/IngesterRunner.js
```diff
@@ -0,0 +1,163 @@
...
+      this.retryHandleFileUploaded(filePath);
+    });
+  }
+
+  retryHandleFileReady(filePath) {
+    const retries = this.retryUploadCounts.get(filePath) || 0;
+
+    if (retries >= this.MAX_RETRIES) {
+      console.error(`RETRYING UPLOAD: Max upload retries reached for ${filePath}. Giving up.`);
+      return;
```
amrc-benmorrow, line 135: Having accepted this file for processing we cannot stop retrying until we succeed. It might be helpful to post an Alert to the Edge Agent if we have files which are repeatedly failing, but we can't stop.

edge-files-tdms/lib/IngesterRunner.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    this.retryUploadCounts.set(filePath, retries + 1);
+    console.warn(`RETRYING UPLOAD: Attempt ${retries + 1} for ${filePath}, retrying in 5 seconds...`);
+
+    setTimeout(() => {
+      this.eventManager.emit(EVENTS.FILE_READY, { filePath });
+    }, 5000);
```
amrc-benmorrow, line 143: Retry delays like this should always have some random jitter in them; that is, something like `5000 + Math.random() * 2000`. This avoids unexpected synchronisation issues when something goes wrong and you get many files retrying all at once.

Also I think this is much too short. Any error here is likely to need a human admin to resolve, so I would delay a minimum of 5 minutes.

amrc-benmorrow (commented):
This looks like you're trying to be careful and handle errors properly and so on, which is good. There are a few places where the error handling will mask problems rather than dealing with them; it's important not to catch an exception (and not rethrow) unless you can actually deal with the problem.

Many of these comments are just general style questions and don't necessarily need addressing on this project, but I think it's worth being aware of them anyway.

edge-files-tdms/bin/api.js
```diff
@@ -0,0 +1,58 @@
...
+  STATE_FILE: env.STATE_FILE,
+  TDMS_DIR_TO_WATCH: env.TDMS_DIR_TO_WATCH,
+  NODE_ENV: env.NODE_ENV,
+  PYTHON_SUMMARISER_SCRIPT: env.PYTHON_SUMMARISER_SCRIPT,
+}).init();
+
+api.run();
+
+await ingesterRunner.run()
+.catch(err => {
```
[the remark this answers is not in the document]
amrc-za: Removed the catch.

edge-files-tdms/lib/IngesterRunner.js
```diff
@@ -0,0 +1,163 @@
...
+
+    this.FailureEvents = [
+      EVENTS.FILE_UPLOAD_FAILED,
+      EVENTS.FILE_ADD_AS_CLASS_MEMBER_FAILED,
+      EVENTS.FILE_UUID_FAILED,
+      EVENTS.FILE_READY_FAILED,
+    ];
+  }
+
+  init(){
```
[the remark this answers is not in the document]
amrc-za: Removed init, moved contents into the constructor

edge-files-tdms/lib/tdms-file-events.js
```diff
@@ -0,0 +1,25 @@
+import { EventEmitter } from 'events';
+
+export const EVENTS = {
```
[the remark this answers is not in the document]
amrc-za: Removed the EVENTS constants and replaced with strings.

edge-files-tdms/lib/api/api-v1.js
```diff
@@ -82,7 +93,9 @@ export class APIv1 {
 
             this.log("Renaming file.");
             // Rename the temporary file now it's written to disk.
-            await fs.rename(tempPath, filePath);
+            await fs.rename(tempPath, finalPath);
+            this.eventManager.emit('file:detected', { filePath: finalPath });
```
amrc-benmorrow, line 97: You need to think about whether this `emit` should come before or after the response is sent. Node events are synchronous which means that all listeners will have been invoked before `emit` returns, and any exceptions thrown will cause a 500 return from the HTTP call. However this only applies up until the first `await` in each handler; `emit` cannot await a Promise.

It is important that the state file in the StateManager is written successfully before we return 201. It would be better to call directly into the StateManager and await the result rather than relying on events here.

amrc-za: Passing StateManager to the APIv1

edge-files-tdms/lib/uploader.js
```diff
@@ -0,0 +1,129 @@
...
+
+    async processFile(filePath) {
+        console.log(`UPLOADER: Handling new file to upload: ${filePath}.`);
+
+        try {
+            if (!(await waitForFileExist(filePath))) {
+                console.error(`UPLOADER: Filepath ${filePath} does not exist.`);
+                this.eventManager.emit('file:readyFailed', { filePath, error: 'File does not exist' });
+                return;
+            }
```
amrc-benmorrow, line 50: This does not do anything useful. The file should already exist for the event to be emitted. If somehow it didn't this is too early to catch the error; an exception will be thrown by the file upload method when it tries to open a file which isn't there.

edge-files-tdms/lib/uploader.js
```diff
@@ -0,0 +1,129 @@
...
+            if (fileState?.isUploaded) {
+                console.log(`UPLOADER: Skipping ${filePath}, already uploaded.`);
+                return;
+            }
+
+            let fileUuid = fileState?.uuid;
+
+            if (!fileUuid) {
+                fileUuid = await this.createFileUuid(filePath);
+                if (!fileUuid) return;
```
amrc-benmorrow, line 63: If any of these steps fail we need to retry. It's probably best to throw an exception and let the retry logic do its job.

edge-files-tdms/lib/uploader.js
```diff
@@ -0,0 +1,129 @@
...
+            }
+            console.error(`UPLOADER: Error uploading ${filePath}`, fsError);
+            this.eventManager.emit('file:uploadFailed', { filePath, error: fsError });
+            return false;
+        }
+    }
+
+    async addAsClassMember(fileUuid, filePath) {
+        try {
+            await this.fplus.ConfigDB.class_add_member(File_Type.TDMS, fileUuid);
```
amrc-benmorrow, line 118: This class UUID should not be hardcoded. In general this will want to be a 'TDMS files from &lt;device>' subclass of the TDMS file type class so it needs to be configurable.

amrc-benmorrow (changes requested):

edge-files-tdms/lib/uploader.js
amrc-za: Removed the hardcoded uuids, passing as arguments.

amrc-benmorrow (changes requested):

edge-files-tdms/lib/api/api-v1.js
```diff
@@ -109,14 +120,23 @@ export class APIv1 {
                         message: 'File uploaded successfully to the driver.',
                         filename: filename,
                     });
-                }catch(err){
+                } catch (err) {
+                    await cleanupFinalPath();
                     console.error(`Failed to persist state for file ${finalPath}:`, err);
-                    return res.status(500).json({error: err.message});
+                    return res.status(500).json({ error: err.message });
```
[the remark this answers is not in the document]
amrc-benmorrow: This also does not need changing now, this is just for future development.

edge-files-tdms/bin/api.js
```diff
@@ -51,7 +52,9 @@ const ingesterRunner = new IngesterRunner({
   fplus: fplus,
   env: env,
   eventManager: eventManager,
-  stateManager: stateManager
+  stateManager: stateManager,
+  fileTypeClass: File_Type.TDMS,
+  fileClass: Class.File,
```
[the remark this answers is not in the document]
amrc-za: Added as env var

amrc-benmorrow:
This is not going to be merged to ACS.

PR #545, opened 2025-07-21
Support notify on the Auth service

amrc-benmorrow:
Currently this supports these requests:

    WATCH v2/principal/
    WATCH v2/identity/kerberos/
    WATCH v2/acl/:principal
    WATCH v2/acl/kerberos/:upn

This is sufficient to support a service to track its ACLs, but not sufficient to support the Auth editor this would ideally need SEARCH on both identities and grants.

This involves a simplifying the Notify class in `service-api`. Instead of subclassing this class, this now acts like an Express router with callbacks to provide notify sequences. This is an incompatible change but this API isn't in a published version of the module yet.

### DIFF
--- a/acs-configdb/lib/routes.js
+++ b/acs-configdb/lib/routes.js
@@ -1,7 +1,5 @@
 /*
- * ACS ConfigDB
- * Express routing
- * Copyright 2024 University of Sheffield AMRC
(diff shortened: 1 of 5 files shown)


djnewbould (approved):
Looks good. Tested and works as expected.

PR #555, opened 2025-08-13
Decrease minimum Grafana dashboard refresh interval

KavanPrice:
Adds a minimum refresh interval to `grafana` in `values.yaml`.

This gives people a bit more freedom in setting their Grafana dashboard refresh rates. The 500ms interval is a _minimum_ and is set in the dashboard settings; the _default_ interval remains 5s.

### DIFF
--- a/deploy/values.yaml
+++ b/deploy/values.yaml
@@ -342,6 +342,8 @@ grafana:
       header_name: X-Auth-Principal
       header_property: username
       auto_sign_up: true
+    dashboards:
+      min_refresh_interval: 500ms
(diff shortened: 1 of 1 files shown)


AlexGodbehere:
I think it would be worth clarifying the problem we are trying to address here. Reducing the minimum refresh interval could place unnecessary load on the InfluxDB instance without delivering real benefits. Since data points are only flushed to InfluxDB every 10 seconds, I’m not sure this change would achieve the intended outcome unless I’m misunderstanding the requirement.

PR #559, opened 2025-09-02
Granular ConfigDB object permissions

amrc-benmorrow:
**This has not been fully tested and is not ready for merge. Also it should be rebased onto main.**

Currently the _Manage objects_ ConfigDB permission covers a large range of actions. This is because the object model has been made richer but the permission structure has not been expanded to cover new actions.

Divide the _Manage objects_ permission into finer-grained permissions. Ensure appropriate permissions are granted to appropriate services. This requires use of ownership and _Mine_ permissions, so a fixup step is needed to reset ownership on some existing objects.

### DIFF
--- a/acs-auth/lib/api_v2.js
+++ b/acs-auth/lib/api_v2.js
@@ -142,6 +142,11 @@ export class APIv2 {
 
         const grant = req.method == "PUT" ? req.body : null;
         if (grant && !valid_grant(grant)) fail(422);
+        /* XXX There is a race condition here but I can't see how to
+         * avoid it. Our permitted list is built here, before the txn,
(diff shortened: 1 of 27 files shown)


amrc-benmorrow:
This cause creation of a new cluster to take over half an hour to reconcile. This is because `sv1clustermanager` now has _Write class members_ on _Clusters-managed groups_ rather than _Manage objects_ on wildcard. The newly-created edge flux / edge krbkeys objects are created in the appropriate clusters-managed group, but the ConfigDB auth caching means that 403 is returned on the initial attempt to set class membership. When the cluster manager retries in 10 minutes one step succeeds but then the next fails. I think #560 is a necessary part of the solution here.

PR #562, opened 2025-09-04
Rtde driver

Harrier-SA:
A driver for connecting to devices using rtde. Primarily Universal Robots.


KavanPrice:
My suggestion here would be to pass the addresses straight through and ignore them within the driver. The grabbing of metric values is then done purely through JSON paths. A user of this driver should then input all metric addresses as `/` so the Edge Agent can optimise its address logic. From the driver's perspective, it gets given one meaningless address, `/`, and blindly returns whatever it receives on its TCP socket against this address.

edge-rtde/lib/rtde.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  async close() {
+    return new Promise((resolve) => {
+      this.client.once("close", () => {
+        this.log("RTDE connection closed.");
+        resolve();
+      });
+
+      this.client.end();
```
amrc-benmorrow: The call to `client.end()` never happens; the Promise is returned immediately. The library carefully supports dual callback and Promise APIs to avoid the need for this :).
```
close(done) {
    this.client.once("close", done);
    this.client.end();
}
```

It's important in this case that `close` itself returns `undefined`; specifically it mustn't be `async` as then it returns a Promise.

lib/js-edge-driver/lib/driver.js
```diff
@@ -279,9 +279,15 @@ export class Driver {
...
 
         this.message("addr", async p => {
+            //this.log("📨 Received ADDR message: %s", p.toString());
             const a = this.json(p);
-            if (!a || !await this.setAddrs(a))
-                this.setStatus("ADDR");
+            //this.log("📋 Parsed addrs: %O", a);
+            if (!a || !await this.setAddrs(a)) {
+                //this.log("❌ Failed to set addrs");
+                //this.setStatus("ADDR");
```
amrc-benmorrow: This call to `setStatus` is important; it informs the Edge Agent that an invalid address has been configured. I think removing this may be a consequence of the confusion around `subscribe` and trying to allow the Edge Agent to define the address?

amrc-benmorrow (approved):
Looks good to me :+1:.

PR #584, opened 2026-01-14
Enhance JSON handling for metrics: stringify objects for string metri…

Harrier-SA:
…cs and parse buffers in MQTT messages
This pull request improves how object values are handled and serialized when dealing with string metrics and MQTT messages. The main focus is to ensure that if a metric expects a string but receives an object, the object is safely stringified to prevent errors and improve data consistency.

**Object-to-string serialization for string metrics:**

* In `acs-edge/lib/device.ts`, when a metric's type is string and the value is an object, the code now attempts to stringify the object. This avoids issues when non-string values are assigned to string metrics.
* In `acs-edge/lib/helpers/typeHandler.ts`, the `parseValueFromPayload` function now stringifies the payload if the metric type is string and the payload is an object, ensuring consistent handling of object-to-string conversions.

**MQTT message parsing improvements:**

* In `acs-edge/lib/devices/MQTT.ts`, the MQTT message handler now tries to parse incoming buffer messages as JSON first, and if that fails, retains the buffer. This makes message processing more robust and flexible for different message formats.

### DIFF
--- a/acs-edge/lib/device.ts
+++ b/acs-edge/lib/device.ts
@@ -400,13 +400,11 @@ export class Device {
                                 this._delimiter
                             ) : obj[addr];
 
-                            // If metric type is String and newVal is an object, stringify it
+                            // Detect misconfiguration: String metric received an object
                             if (metric.type === sparkplugDataType.string && typeof newVal === 'object' && newVal !== null && !Buffer.isBuffer(newVal)) {
-                                try {
-                                    newVal = JSON.stringify(newVal);
-                                } catch (e) {
-                                    log(`ERROR - Stringify failure: ${e}`);
-                                }
+                                log(`⚠️ ERROR - Misconfiguration: Metric '${metric.name}' expects String but received object. Check payloadFormat and path config.`);
+                                log(`   Object received: ${JSON.stringify(newVal).substring(0, 100)}...`);
+                                newVal = null; // Publish null to indicate error
                             }
 
                             // Test if the value is a bigint and convert it to a Long. This is a hack to ensure that the
--- a/acs-edge/lib/helpers/typeHandler.ts
+++ b/acs-edge/lib/helpers/typeHandler.ts
@@ -500,14 +500,6 @@ export function parseValueFromPayload(msg: any, metric: sparkplugMetric, payload
                 } else if (typeof payload === 'object' && payload.hasOwnProperty('value')) {
                     // If no path is provided but the object has a 'value' property, use that
                     return parseTypeFromString(metric.type, payload.value);
-                } else if (metric.type === sparkplugDataType.string && typeof payload === 'object' && payload !== null) {
-                    // If metric type is String and payload is an object (no path), stringify it
-                    try {
-                        return JSON.stringify(payload);
-                    } catch (e) {
-                        log(`ERROR - Stringify failure: ${e}`);
-                        return null;
-                    }
                 }
             }
             break;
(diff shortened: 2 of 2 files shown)


acs-edge/lib/devices/MQTT.ts
```diff
@@ -67,7 +67,21 @@ export class MQTTConnection extends DeviceConnection {
...
+                } catch (e) {
+                    // If not JSON, keep as buffer
+                    obj[topic] = msg;
+                }
+            } else if (typeof msg === 'object' && msg !== null) {
+                // If it's already an object, keep it
+                obj[topic] = msg;
+            } else {
+                obj[topic] = msg;
+            }
```
AlexGodbehere: It's been a while since I looked at the edge-agent code in any amount of detail, but I _think_ devices should only ever emit buffers. @amrc-benmorrow can you confirm this?

@Harrier-SA does this actually need to be in here or are the other two changes sufficient?

acs-edge/lib/helpers/typeHandler.ts
```diff
@@ -500,6 +500,14 @@ export function parseValueFromPayload(msg: any, metric: sparkplugMetric, payload
                 } else if (typeof payload === 'object' && payload.hasOwnProperty('value')) {
                     // If no path is provided but the object has a 'value' property, use that
                     return parseTypeFromString(metric.type, payload.value);
+                } else if (metric.type === sparkplugDataType.string && typeof payload === 'object' && payload !== null) {
+                    // If metric type is String and payload is an object (no path), stringify it
+                    try {
+                        return JSON.stringify(payload);
+                    } catch (e) {
+                        log(`Failed to stringify JSON payload for metric ${metric.name}: ${e}`);
+                        return null;
```
AlexGodbehere: Better to just return something more meaningful here like `ERROR - Stringify failure: ${e}`

AlexGodbehere (commented):
@amrc-benmorrow for context, @Harrier-SA is trying to transmit a JSON object with many, many values as a "blob" through ACS instead of configuring each one. I think it's a valid use case, but would like your input on this.

acs-edge/lib/devices/MQTT.ts
Harrier-SA: @AlexGodbehere  No this is likely an overstep on my side. I have tested my use case without it and it seems to still work okay so will undo these changes.

acs-edge/lib/device.ts
```diff
@@ -400,6 +400,15 @@ export class Device {
                                 this._delimiter
                             ) : obj[addr];
 
+                            // If metric type is String and newVal is an object, stringify it
+                            if (metric.type === sparkplugDataType.string && typeof newVal === 'object' && newVal !== null && !Buffer.isBuffer(newVal)) {
+                                try {
+                                    newVal = JSON.stringify(newVal);
+                                } catch (e) {
+                                    log(`Failed to stringify object for metric ${metric.name}: ${e}`);
```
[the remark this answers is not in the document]
Harrier-SA: Updated as suggested.

amrc-benmorrow:
OK it looks like we can't accept a string via a Buffer payload. Fixing that would be a better way forward here. Probably the `A:B` 'bit' path format wants generalising to support 'start:end' for a string, with an omitted end going to the end of the buffer.

In principle we should consider character sets; Sparkplug Strings are UTF-8 but the source might not be. However just assuming UTF-8 is probably OK for our uses, and is what the current JSON parsing code does.

amrc-benmorrow (approved):
Cool, thanks!

PR #597, opened 2026-02-17
Implement notify support in Java service client

amrc-benmorrow:
Implement the client side of the notify/v2 protocol.

This pulls Jetty for WebSocket support, primarily because it's the only decent implementation I can find. The Jetty server-side stuff is also what Jena Fuseki uses for its SPARQL server. We also pull Vavr as an immutable data structures library, in the absence of anything decent in the Java standard library. We also now require Java 17 for `record` support.

This means we now have two HTTP client libraries. Unfortunately the Jetty client has no client-side cache support; generally it appears to be less complete than the Apache libraries. This means for now we have no choice but to leave the REST support using the Apache library, as HiveMQ currently depends critically on the cache support for performance. If the HiveMQ plugin is ever rewritten to use notify instead of GET then we could probably switch to the Jetty HTTP client.

Implementation largely follows the mould of the JS libraries. The generics are getting a little out of hand in some places but without type aliases I'm not sure what can be done about that: the types in use are genuinely complicated. Some functionality is omitted for now, such as headers on Response objects, filters on SEARCH requests and powerset tracking in the ConfigDB interface; these do not yet have a use case.

### DIFF
--- a/acs-admin/src/App.vue
+++ b/acs-admin/src/App.vue
@@ -1,5 +1,5 @@
 <!--
-  - Copyright (c) University of Sheffield AMRC 2025.
+  - Copyright (c) University of Sheffield AMRC 2026.
   -->
 
 <template>
@@ -73,6 +73,7 @@
         <NewEdgeDeploymentDialog/>
         <NewDeviceDialog/>
         <NewConnectionDialog/>
+        <NewBridgeDialog/>
         <RouterView/>
       </main>
     </SidebarInset>
@@ -97,6 +98,7 @@ import NewClusterDialog from '@components/EdgeManager/EdgeClusters/NewClusterDia
 import NewEdgeDeploymentDialog from '@components/EdgeManager/Nodes/NewEdgeDeploymentDialog.vue'
 import NewConnectionDialog from '@components/EdgeManager/Connections/NewConnectionDialog.vue'
 import NewDeviceDialog from '@components/EdgeManager/Devices/NewDeviceDialog.vue'
+import NewBridgeDialog from '@components/Bridges/NewBridgeDialog.vue'
 import { ACS_VERSION } from '@/lib/version.js'
 
 export default {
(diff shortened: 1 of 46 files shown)


lib/java-service-client/setup.jsh
```diff
@@ -0,0 +1,34 @@
...
+import io.reactivex.rxjava3.core.*;
+import io.reactivex.rxjava3.core.Observable;
+import io.reactivex.rxjava3.core.Observer;
+import io.reactivex.rxjava3.disposables.*;
+import io.reactivex.rxjava3.observers.*;
+
+import uk.co.amrc.factoryplus.client.*;
+import uk.co.amrc.factoryplus.util.*;
+
+void log (String s) { System.out.println(s); }
```
KavanPrice: What's the reason for avoiding `println` directly?

lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/http/WsRequest.java
```diff
@@ -0,0 +1,88 @@
...
+                    final var rv = raw.getReceiver()
+                        .firstOrError()
+                        .flatMap(status -> {
+                            log.info("WS auth response: {}", status);
+
+                            if (status.equals("200"))
+                                return Single.just(raw);
+
+                            if (status.equals("401"))
+                                throw new BadToken(base, token);
```
KavanPrice: Avoid magic numbers. I know they're well-known status codes but we're pulling in Apache/Jetty anyway so we should use them.

lib/java-service-client/setup.jsh
amrc-benmorrow: This is a startup script for `jshell`, the Java REPL, it's not part of the build. In the REPL you can't access `println` directly, you have to type the full `System.out.println`, which is just really annoying.

PR #606, opened 2026-03-11
Data access dumps

amrc-za:


acs-service-setup/dumps/data-access.yaml
```diff
@@ -0,0 +1,141 @@
+
+service: !u UUIDs.Service.ConfigDB
+version: 1
+objects:
+
+  !u UUIDs.Class.Service:
+    !u DataAccess.Service.DataAccess:
+      name: "Data Access Service"
+    
+  !u UUIDs.ConfigDB.R1Class:
```
amrc-benmorrow: The names under `UUIDs` are the names defined in `lib/js-service-client/lib/uuids.js`. The names defined in `acs-service-setup/lib/uuids.js` do not have a prefix.

These UUID names are currently a bit of a mess. It might be worth tidying them up at some point, and possible just moving the whole lot into js-service-client, or into a new `@amrc-factoryplus/uuids` library.

amrc-benmorrow (changes requested):
This is pretty good; most of the structure is correct. There are a few systematic problems where you haven't understood the dump format correctly, but although I think I remember documenting it at some point I honestly couldn't tell you where that document is, so you must have been purely working from the examples.

You obviously haven't tried to build the image or load it into an ACS instance, or some of these problems would have been more obvious. We need to go over how to do that.

You've missed out the `MES` class 🙃. (Though strictly speaking this might be part of Kavan's work rather than yours.)

PR #609, opened 2026-03-20
Begin implementation of RDF ConfigDB

amrc-benmorrow:
Begin implementation of an RDF-backed replacement for the ConfigDB. This is called the Metadata Database, or MetaDB.

Currently this implements most of (the useful parts of) the ConfigDB HTTP API. It also implements `v2/sparql` and `v2/rdf` endpoints for SPARQL queries and Graph Store operations respectively. The `v1` and other compat endpoints are not implemented, including in particular `v1/search` which has no functional replacement.

Not implemented yet:
* Auth of any kind. All operations are permitted.
* Dump loading.
* Notify.
* Useful mappings between RDF and JSON.

### DIFF
--- a/acs-metadb/src/main/java/uk/co/amrc/factoryplus/metadb/api/Sparql.java
+++ b/acs-metadb/src/main/java/uk/co/amrc/factoryplus/metadb/api/Sparql.java
@@ -166,17 +166,17 @@ public class Sparql {
     private Model resolveGraph (boolean create)
     {
         var dataset = store.dataset();
-        var headers = uriInfo.getQueryParameters();
-        var isDef = headers.getFirst("default");
-        var graph = headers.getFirst("graph");
+        var params = uriInfo.getQueryParameters();
+        var isDef = params.getFirst("default");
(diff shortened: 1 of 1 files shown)


acs-metadb/ttl/core.ttl
```diff
@@ -0,0 +1,172 @@
...
+    :uuid "04a1c90d-2295-4cbe-b33a-74eded62cbf1";
+    :rank "1"^^xsd:int; :powersetOf :Individual.
+:R2Class a :R3Class; :primary :R3Class;
+    rdfs:subClassOf :Class;
+    :uuid "705888ce-53fa-434d-afee-274b331d4642";
+    :rank "2"^^xsd:int; :powersetOf :R1Class.
+# The top rank class has no primary class
+:R3Class a :Class; rdfs:subClassOf :Class;
+    :uuid "52b80183-6998-4bf9-9b30-132755e7dede";
+    :rank "3"^^xsd:int; :powersetof :R2Class.
```
KavanPrice: Typo, should be `:powersetOf`

PR #629, opened 2026-04-09
Implement change-notify on the MetaDB

amrc-benmorrow:
Only WATCH and SEARCH on `v2/app/{app}/object/` are implemented for now as we don't yet have a mechanism for tracking the class structure changes.

Implement App config schema validation. Originally this was going to use the notify interface internally but it turned out to be necessary to update the schemas synchronously so we can produce appropriate 409 response when existing entries don't validate. This means a SPARQL update can insert invalid entries but that is currently a superuser facility anyway.

Replace the JsonValueBodyReader from Jakarta EE. It has an important bug where it cannot decode a single-character JSON document, such as `1`. This is because of instead of handling charsets correctly it tries to guess a Unicode charset.


acs-metadb/src/main/java/uk/co/amrc/factoryplus/notify/SearchProvider.java
KavanPrice, line 1: This is futureproofing I assume. I don't see it used anywhere.

KavanPrice (dismissed):

acs-metadb/src/main/java/uk/co/amrc/factoryplus/notify/SearchProvider.java
amrc-benmorrow: No, this is redundant. I'll remove it.

The JS model I was following uses a structure like this, but for this version I ended up not using it because of the need to ensure all requests fall inside the correct transaction.

PR #631, opened 2026-04-14
Wire up MetaDB to the build

amrc-benmorrow:
This requires quite a bit of refactoring of the build of the other Java components. For local builds we can deploy the Maven packages to an in-tree repo and rely on picking them up again from there; for GHA builds the checkout directory does not persist between build steps so we need to push to the GH Packages Maven repo instead. Refactor all the Java POMs to inherit from a new `base-pom`, which handles switching between repos and also unifies the library versions in use.

The Helm deployment allows for the MetaDB to be optionally deployed alongside the ConfigDB, or replacing it. Currently ACS will not function correctly with the MetaDB as the ConfigDB as more work is needed to implement all the endpoints.


deploy/templates/metadb/metadb.yaml
```diff
@@ -0,0 +1,116 @@
...
+              value: sv1metadb@{{ .Values.identity.realm }}
+            - name: SERVER_KEYTAB
+              value: /keytabs/server
+            # This needs to name a principal available in the keytab, used for verifying passwords.
+            - name: SERVER_PRINCIPAL
+              value: HTTP/metadb.{{ .Release.Namespace }}.svc.cluster.local
+            - name: DIRECTORY_URL
+              value: http://directory.{{ .Release.Namespace }}.svc.cluster.local
+            - name: VERBOSE
+              value: {{.Values.mqtt.verbosity | quote | required "values.mqtt.verbosity is required!"}}
```
KavanPrice: Assume this is supposed to be `.Values.metadb.verbosity`?

.github/workflows/publish.yml
```diff
@@ -480,7 +532,7 @@ jobs:
           tags: ghcr.io/amrc-factoryplus/${{ steps.prepare.outputs.service-name }}:${{ steps.prepare.outputs.version }}
           labels: ${{ steps.meta.outputs.labels }}
```
KavanPrice: There is no step with ID `meta`. Everything else uses `prepare`.

KavanPrice (changes requested):

deploy/templates/metadb/metadb.yaml
amrc-benmorrow: :+1:

.github/workflows/publish.yml
amrc-benmorrow: Interesting… I didn't change that. But yes, I agree it looks wrong.

KavanPrice (approved):
LGTM

PR #634, opened 2026-04-17
Auth support for the MetaDB

amrc-benmorrow:
Support authentication and authorisation in the MetaDB. Use the service client to authenticate against Kerberos and to check ACLs. Use the notify framework from the start so ACL changes are pushed to us.

Unlike the ConfigDB implementation, this version handles dynamic ACL changes correctly. A notify client whose permissions change will be correctly notified of their current situation.


amrc-benmorrow:
I'm merging this, I need to move forward. @KavanPrice feel free to add review comments if you like, I'll look at them in a new PR.

PR #635, opened 2026-04-20
Za/data access register with directory

amrc-za:


acs-service-setup/dumps/data-access.yaml
```diff
@@ -1,4 +1,4 @@
-
+#-REQUIRE: auth directory
```
amrc-benmorrow: You need `auth` here but I don't think you need `directory`. These names refer to dump files, and `directory.yaml` is all stuff internal to the Directory. The `service-urls.yaml` which registers the services is always loaded before everything else in any case (it has `#-EARLY`) because we can't even find the servies to load into before that's loaded.

amrc-benmorrow (changes requested):

acs-service-setup/dumps/data-access.yaml
amrc-za: removed directory

amrc-benmorrow (approved):

PR #637, opened 2026-04-22
Start integrating the MetaDB as the ConfigDB

### LINKED ISSUE #567: Admin UI should include a switch to turn on verbose logging
Currently the only way to get verbose logging from the admin UI is to use a local build. Ideally we should have a switch to turn it on which is sticky across sessions so you can get logs from the start of the session.

amrc-benmorrow:
This does not yet get to the point where ACS is functional, but it's a start. The most important missing feature is the MetaDB cannot yet load ConfigDB dumps; there is a dummy endpoint that returns 204, so that service-setup doesn't fall over, but it doesn't load the data.

The admin UI is functional and can be used to view the MetaDB entries. The other services are correctly contacting the MetaDB; mostly they are getting 403s as the accounts and roles are not loaded in correctly yet. Bootstrap ACL support is implemented to allow the Auth service to contact the MetaDB; without this we have a dependency loop and both services hang.

Important changes:
- Support GSSAPI auth in the service-api.
- Support bootstrap ACLs in the Java service-client and use them in the MetaDB.
- Additional HTTP and notify endpoints as needed to make things work.
- CORS support, needed for the admin UI and any other web UI.
- Verbose logging option in the admin UI.
- Create the MetaDB service account using the Local UUID mechanism rather than via krbkeys.
- Move ACL lookups out of the Jena transactions to avoid txns staying open too long.

Closes: #567

### DIFF
--- a/acs-metadb/src/main/java/uk/co/amrc/factoryplus/providers/AuthProvider.java
+++ b/acs-metadb/src/main/java/uk/co/amrc/factoryplus/providers/AuthProvider.java
@@ -15,13 +15,14 @@ import java.time.Instant;
 import java.util.Base64;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.core.*;
 import io.vavr.control.Option;
 
 import uk.co.amrc.factoryplus.client.FPServiceClient;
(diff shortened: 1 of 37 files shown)


acs-metadb/src/main/java/uk/co/amrc/factoryplus/providers/AuthProvider.java
KavanPrice, line 156: This is the only place where sessions are removed. If a client ends their session and doesn't try to use their expired token again I think this will memory leak.

deploy/templates/metadb/metadb.yaml
```diff
@@ -69,13 +69,17 @@ spec:
...
               value: /keytabs/server
-            # This needs to name a principal available in the keytab, used for verifying passwords.
+            # This needs to name a principal available in the keytab, 
+            # used for verifying passwords.
             - name: SERVER_PRINCIPAL
               value: HTTP/metadb.{{ .Release.Namespace }}.svc.cluster.local
             - name: DIRECTORY_URL
               value: http://directory.{{ .Release.Namespace }}.svc.cluster.local
-            - name: VERBOSE
-              value: {{.Values.metadb.verbosity | quote | required "values.metadb.verbosity is required!"}}
```
KavanPrice, line 78: Was `VERBOSE` removed here intentionally?

KavanPrice (changes requested):

deploy/templates/metadb/metadb.yaml
amrc-benmorrow: Yes, it's not used by the Java service-client. Logging levels would need to be set using slf4j Java system properties.

acs-metadb/src/main/java/uk/co/amrc/factoryplus/providers/AuthProvider.java
amrc-benmorrow: Nice catch. I think the js-service-api has this bug too…

KavanPrice (approved):
LGTM

PR #639, opened 2026-04-27
lib rx-client configdb.js - added watch members subclasses method

amrc-za:


lib/js-rx-client/lib/configdb.js
```diff
@@ -127,6 +127,17 @@ export class ConfigDB extends Interfaces.ConfigDB {
...
     }
 
+    /** Watch the subclasses of all members of a class.
+     *
+     * @returns A sequence of immutable Maps from UUID to Set.
+     * @param klass A class UUID.
+     */
+    watch_members_direct_subclasses(klass){
+        return this.watch_members(klass)
+            .pipe(this.expand_direct_subclasses());
```
amrc-benmorrow, line 158: I think for what you are doing you want all subclasses, direct as well as indirect. This means an `expand_subclasses` method which doesn't exist yet but has the same shape as the others.

PR #642, opened 2026-04-29
Load ConfigDB dumps into the MetaDB

amrc-benmorrow:
This includes a number of fixes to the existing dumps to make them load correctly, and to make them compatible with the MetaDB.

* Remove all support for version 1 ConfigDB dumps and upgrades all dumps in ACS to version 2.
* Support `iri` and `powersetOf` keys for an object definition. The ConfigDB accepts these but ignores them. The MetaDB records powerset information but currently does nothing with it.
* Promote _Application_ to R2 and include a new R1 class, _Config entry_, representing individual config entries. The MetaDB treats these as F+ objects with UUIDs. (The ConfigDB does not.)
* Follow through making all Apps subclasses of _Config entry_.
* Revert treating config entries as F+ objects due to Jena performance problems. Leave the class structure in place for future work.

Service-setup now gets to the point of loading all the dumps. It fails immediately afterwards because of the missing `v1/search` endpoint.


acs-metadb/src/main/java/uk/co/amrc/factoryplus/metadb/db/ZFC.java
```diff
@@ -0,0 +1,153 @@
...
+            filter (?rank > 0)
+            filter not exists { ?obj rdfs:subClassOf <core/Object>. }
+        }
+    """);
+    private static final Query V_indSubclass = Vocab.query("""
+        select ?objU
+        where {
+            ?objU a <core/Individual>; rdfs:subClassOf ?class.
+        }
+    """);
```
KavanPrice, line 99: Should this have `?obj <core/uuid> ?objU;` as with the rest of these queries?

acs-metadb/src/main/java/uk/co/amrc/factoryplus/metadb/db/RdfStore.java
```diff
@@ -241,6 +252,43 @@ public void runUpdate (UpdateRequest update, Object... substs)
...
+        var ttl = RdfStore.class.getResourceAsStream("core.ttl");
+        RDFDataMgr.read(core, ttl, Lang.TURTLE);
+
+        var coreVer = findVersion(core).get();
+
+        this.executeWrite(() -> {
+            findVersion(direct)
+                .peek(ver -> {
+                    log.info("Found core schema version {}", ver);
+                    if (ver != coreVer)
```
KavanPrice: I think this is another annoying reference comparison in Java where you actually want a value comparison. i.e. `!ver.equals(coreVer)`

KavanPrice (changes requested):

acs-metadb/src/main/java/uk/co/amrc/factoryplus/metadb/db/ZFC.java
amrc-benmorrow: Ooh, that's just wrong. Yes, it needs fixing.

acs-metadb/src/main/java/uk/co/amrc/factoryplus/metadb/db/RdfStore.java
amrc-benmorrow: OK, question for you, since I think you know more Java than me: does changing the type of `coreVer` to be explicitly `int` rather than leaving it to default to `Integer` fix this? That is, if we have
```
int coreVer = 1;
Integer ver = Integer.valueOf(1);
```
then does `ver == coreVer` (which does compile, I've checked) cast `ver` to `int` and do a primitive comparison or cast `coreVer` to `Integer` and do a reference comparison? (Which will definitely succeed, by spec, until the version number goes above 127…)

C#'s generics-over-value-types cause a lot of trouble with Unity and AOT compilation, but they do simplify this sort of situation…

KavanPrice: They both get promoted to primitives: https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.21.1

This extends to promotable types in general. Personally I think hiding things like this is unacceptable language design, but I'm also OTT.

KavanPrice (approved):
LGTM

PR #643, opened 2026-04-30
Za/data access webapi

amrc-za:
Implemented web api feature. Includes: 
- dataflow
- api-v1
- lib/js-service-client/lib/data-access.js client interface for data access service


acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+import * as rx from "rxjs";
+
+import { APIError } from "@amrc-factoryplus/service-api";
+import { retryBackoff } from "@amrc-factoryplus/rx-util";
+import { DataAccess as Constants } from "./constants.js";
+import { valid_uuid, DatasetValidity } from "./validate.js";
+import {convert_to_csv, minDate, maxDate} from './utils.js';
+
+function fail(status, message) {
+  throw new APIError(status);
```
amrc-benmorrow: This does not include the message anywhere… logging it would be OK if there's nothing better to do.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+  */
+  async metadata_list(req, res) {
+    const uuids = await this.data.get_allowed_dataset_uuids(req.auth, Constants.Perm.ReadDataset, DatasetValidity.VALID);
+    return res.status(200).json(uuids);
+  }
+
+
+  async _get_dataset_time_bounds(dataset_uuid, visited=new Set()){
+
+    if(visited.has(dataset_uuid)) return fail(422, `Cycle detected at dataset ${dataset_uuid}`);
```
amrc-benmorrow: This is called from read endpoints such as GET /metadata/:uuid. These cannot return a 422 status, this applies to the request body, and a GET has no request body. Instead they should treat this dataset as invalid.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+        to: minDate(to, child.to)
+      };
+    }
+
+
+    // ------------------------
+    // 2. SPARKPLUG SRC
+    // ------------------------
+    if(structure === Constants.App.SparkplugSrc){
+      const {from, to} = await this.influxReader.get_influx_time_bounds(config.source);
```
amrc-benmorrow: These from/to bounds are not supposed to be looking at Influx at all; they are structural bounds set by the Sessions. A SparkplugSrc has null bounds.

If it's easy to extract real first/last bounds on all datasets from Influx that might be a useful addition to the API, but it would want to be returned via different properties. It may even be worth having a dedicated endpoint for that information, if it's useful, because it will change much faster than the structural information. Every time a new Sparkplug packet comes in the 'last timestamp in Influx' is potentially updated.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+
+    }else{
+      // Unhandled Structure type
+      return fail(422, `Structure ${structure} is unknown`);
+    }
+  }
+
+
+  _merge_intervals(queries){
+    const grouped = {};
```
amrc-benmorrow: It's better practice in modern JS to use a Map rather than using an object as a dictionary. I would have used an `immutable.Seq` and .groupBy here.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+
+    const result = []; 
+
+    for(const source in grouped){
+        const intervals = grouped[source]
+            .map(q => ({
+                from: q.from ? new Date(q.from) : new Date(0),
+                to: q.to ? new Date(q.to) : new Date()
+            }))
+            .sort((a, b) => a.from - b.from);
```
amrc-benmorrow: You need to be careful in general about using the .sort method on arrays: it sorts in place. In this case the array being sorted is the array just constructed by .map so it isn't a problem, but in general it's safer to use .toSorted which returns a new array.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+      }
+
+      visited.add(dataset_uuid);
+
+      const dataset_def = await this.data.get_dataset_def_by_uuid(dataset_uuid);
+      if (!dataset_def) return fail(404, `Dataset definition not found for ${dataset_uuid}`);
+
+      const { structure, config } = dataset_def;
+
+      if(!structure) return fail(422, `Structure is not found in dataset definition ${dataset_uuid}`);
```
amrc-benmorrow: Although this is part of a POST handler, so we have a request body, this is not reporting a problem with the request body. The problem is with the data in the ConfigDB, which is not the client's responsibility. These should all return 404: a dataset with an invalid definition doesn't exist as far as read-only clients are concerned.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+      Constants.Perm.ReadDataset,
+      dataset_uuid,
+      true,
+    );
+    if (!ok) return fail(403, `You don't have Read permission for dataset ${dataset_uuid}`);
+
+    const dataset_def = await this.data.get_dataset_def_by_uuid(dataset_uuid);
+    if (!dataset_def) return fail(404, `Dataset def ${dataset_uuid} not found or invalid.`);
+
+    const ok2 = await this._check_second_level_permission(
```
amrc-benmorrow: You don't want this permission check. The whole point here is that creating a dataset and giving someone read access to that dataset lets them read the data in the dataset even if they can't read the source data.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+      const results = await Promise.all(
+        merged.map(q => 
+            this.influxReader.get_dataset_data(q.source, {
+                from: q.from?.toISOString(),
+                to: q.to?.toISOString()
+            })
+        )
+      );
+
+      const rows = results.flat();
```
amrc-benmorrow: This holds all the results from Influx in memory, and converts to CSV in one go. This will need some careful testing to make sure the memory requirements don't get silly when realistic datasets are downloaded.

The alternative would be to stream the data via a CSV converter. Where a dataset consists of multiple parts it would be necessary to stream them one after another, so the data would come out in the CSV out of sequence; I don't think that should be a problem as everything's timestamped.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+      this.data.get_dataset_definitions(DatasetValidity.ALL).pipe(
+        rx.map(datasets => datasets.get(uuid))
+      )
+    );
+
+    if (!definition) return fail(404, `Dataset def with uuid ${uuid} not found.`);
+
+    if (definition.length > 1){
+      for(let def of definition){
+        def.structure = Constants.Special.InvalidDataset;
```
amrc-benmorrow: Here you are modifying the data structure returned from the sequence. It's very important not to do this, it will cause chaos, because it will change the results seen by other consumers of the sequence. If you were using an immutable List rather than a plain JS Array it would have forced you to make a copy and avoided this; this is why I use them.

In fact I don't think this logic should be here at all; see the comment to _build_dataset_definitions.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+      }
+    }
+
+    return res.status(200).json(definition);
```
amrc-benmorrow: If `definition` is an Array this will return the wrong thing.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    const ok = await this.auth.check_acl(
+      principal,
+      Constants.Perm.CreateDataset,
+      structure,
+      true
+    );
+
+    if (!ok) return fail(403, `You don't have Create permission for structure ${structure}`);
+
+    if(!config) return fail(422, `Config not provided.`);
```
amrc-benmorrow: I would do the basic validation checks that don't look at any existing data before checking permissions. If the request doesn't contain a config it can't ever succeed, it doesn't matter what the permissions are.

It is important to check ACLs before looking at any existing data, to avoid leaking information.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    
+    // Check the principal has appropriate permission for all dataset sources in config. 
+    const ok2 = await this._check_second_level_permission(principal, structure, config, mapStructureToPermission.get(structure));
```
amrc-benmorrow: Now this method is only called here it can map structure UUID to permission inside the check method. Given there's already a switch inside that method the map probably isn't needed at all.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    
+    if(!ok2) return fail(403, `You don't have ${mapStructureToPermission.get(structure)} permission for dataset source(s).`);
+
+    // Create new Dataset Object
+    if(!objectUuid){
+      objectUuid = await this.cdb.create_object(Constants.Class.Dataset);
+      
+      if(!objectUuid) return fail(500, `New Dataset object couldn't be created in ConfigDB.`);
```
amrc-benmorrow: `create_object` cannot return false. If the request to the ConfigDB fails it will throw an exception, which the WebAPI framework will catch and translate into a 503 response, which is the correct response for not being able to access an upstream service.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    return objectUuid;
+  }
+
+  async _create_subclass_relationship(structure, dataset_uuid, config) {
+    if (structure === Constants.App.SessionLimits) {
+
+      await rx.lastValueFrom(
+        rx.defer(() =>
+          this.cdb.class_add_subclass(config.source, dataset_uuid)
+        ).pipe(retryBackoff(500, e => this.log(e)))
```
amrc-benmorrow, line 418: I'm being picky here, but you don't need to duplicate the retry logic.
```
const subclasses = 
    structure == Constants.App.SessionLimits ? [[config.source, dataset_uuid]]
    : structure == Constants.App.UnionComponents ? config.map(source => [dataset_uuid, source])
    : [];
for (const subclass of subclasses) {
    await … class_add_subclass(...subclass) …
}
```
Given the number of places you switch on structure type I would also consider defining a StructureHandler class with subclasses for the different types and methods for all these operations. But that's a larger refactor and probably not worth it now.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    if(structure === Constants.App.SessionLimits){
+      if(curr_config.source != new_config.source){
+        // remove the dataset from subclasses of its source dataset
+        await this.cdb.class_remove_subclass(curr_config.source, dataset_uuid);
+        this.log(`Removed ${dataset_uuid} from ${curr_config.source} subclasses.`)
+      }
+    }else if(structure === Constants.App.UnionComponents){
+      if(curr_config.length > 0){
+        // if new_config does not include sources from curr_config -> remove subclass rel.
+        const sources_to_remove = curr_config.filter(x => !new_config.includes(x));
```
amrc-benmorrow: Given that you will reinstate the correct new subclass relations I wouldn't bother with this step, just remove them all. It's cleaner and it avoids needing to reference the new config from this method.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    const curr_config_for_structure = curr_defs.find(c => c.structure === structure)?.config;
+    
+    if(!curr_config_for_structure){
+      this.log(`Current config for structure does not exist.`);
+    }
+
+    // update subclass links if config def for this structure already exists and being updated.
+    if(curr_config_for_structure){
+      await this._unlink_subclasses(structure, dataset_uuid,curr_config_for_structure, new_config);
+    }
```
amrc-benmorrow, line 531: You are unlinking subclasses based on the **new** structure type. If you allow the structure type to change then this must be performed based on the **current** structure type. Forbidding a change of structure type (with a 409 status) is a valid option here and may be simpler in the long run.

It is important to allow a PUT to a dataset which is currently invalid to succeed. This should not attempt to remove any old subclass relations.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+   */
+  async structure_create(req, res){
+    const objectUuid = await this._update_dataset_config(
+      req.auth,
+      req.body.structure,
+      req.body.config,
+      null
+    );
+
+    if(!objectUuid) return fail(422, `Failed to create new Dataset.`);
```
amrc-benmorrow: This method also cannot return false. If it fails it will throw.

acs-data-access/lib/api-v1.js
```diff
@@ -0,0 +1,701 @@
...
+    if(!objectUuid) return fail(422, `Failed to update dataset ${dataset_uuid}`);
+
+    // remove definitions from other structure apps if the dataset was invalid
+    const invalid_defs = await rx.firstValueFrom(
+
+      this.data.get_dataset_definitions(DatasetValidity.INVALID).pipe(
+        rx.map(datasets => datasets.get(dataset_uuid) ?? []),
+        rx.map(defs => defs.filter(d => d.structure !== req.body.structure)),
+        rx.map(filtered => filtered.length ? filtered : undefined)
+      )
```
amrc-benmorrow: This is all much too clever :). Loop over the structure types, call `delete_config` for each, and catch any 404s. ServiceError.check is useful here. Do this before calling `_update_dataset_config` and you will end up with just the correct new definition.

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+    // this.general_infos.subscribe(x => this.log("General infos UPDATE %o", x.toJS()));
+    // this.parts.subscribe(x => this.log("Parts UPDATE %o", x.toJS()));
+  }
+
+  // watch Dataset's members' subclasses
+  _build_parts() {
+    return rxu.rx(
+      this.cdb.watch_members_subclasses(Constants.Class.Dataset),
+      rx.map(map => map.map(subclass_uuids => 
+          subclass_uuids.filter(x => this.cdb.class_has_member(Constants.Class.Dataset, x))
```
amrc-benmorrow: `class_has_member` returns a Promise, which is a true value, so this filter will not do anything.

You either need a `combineLatest` with `watch_members(Dataset)`, or (better) you want to inline the `switchMap` from `expand_subclasses` and append the filter step there. At that point you will have the list of Datasets in your hand so you just check it.

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+            rx.map(result => [uuid, result])
+          )
+        );
+
+        return rx.combineLatest(streams).pipe(
+          rx.map(entries => {
+            let map = IMap();
+
+            for (const [outerUuid, result] of entries) {
+              for (const [innerUuid, payload] of result) {
```
amrc-benmorrow: This sort of data manipulation can get very confusing. I would call these variables `appUuid` and `objUuid` (or `datasetUuid`) rather than `outer` and `inner`. It makes it clearer what their function is, rather than just their position in the initial data structure. (Well, I would probably call them `app` and `ds`, but that's just me… :) ).

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+      ]),
+
+      rx.switchMap(([sprkDevices, sessions, unions]) => {
+        const grouped = IMap({
+          [Constants.App.SessionLimits]: sessions,
+          [Constants.App.UnionComponents]: unions,
+          [Constants.App.SparkplugSrc]: sprkDevices
+        });
+
+        return rx.from(grouped.entrySeq()).pipe(
```
amrc-benmorrow: This is all synchronous data processing. You can use Rx for this, it won't do any harm, but the immutable library has equivalent methods that are probably easier for synchronous processing.

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+                datasetId,
+                definition: { structure, config }
+              }))
+            )
+          ),
+
+          // Group into: { datasetId: [definitions] }
+          rx.reduce((acc, { datasetId, definition }) => {
+            const existing = acc.get(datasetId, []);
+            return acc.set(datasetId, [...existing, definition]);
```
amrc-benmorrow: I would identify invalid datasets at this point and throw away invalid definitions. If a dataset has exactly one definition, keep it, otherwise just create a new definition `{ structure: StructurallyInvalidDataset }`. Then you don't need to handle a possible array of definitions anywhere else in the code.

It would also be possible to do more detailed validation at this point, including calculating the `from`/`to` bounds. You have the full set of dataset definitions in your hand so checking things like 'do source UUIDs refer to valid datasets' is easy. This check would need to be recursive, and identify self-referencing datasets as invalid, just like your existing from/to logic; but because you're processing the whole collection of datasets at once you can cache the results and only calculate validity and bounds for each dataset once. Then the API calls would just look up the latest results and wouldn't need to make any ConfigDB requests at all. This again may be a larger refactor and not appropriate at this point, but the strategy is worth keeping in mind for the future.

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+    );
+  }
+
+  _build_dataset_definitions() {
+    return rxu.rx(
+      rx.combineLatest([
+        this.cdb.search_app(Constants.App.SparkplugSrc),
+        this.cdb.search_app(Constants.App.SessionLimits),
+        this.cdb.search_app(Constants.App.UnionComponents)
+      ]),
```
amrc-benmorrow: So, you've just written a nice neat function above, `_build_metadata`, that takes a class of Apps and builds a sequence of maps from objects to the configs from that class for those objects. With one change, making the DatasetMetadata class a parameter, you can use it here too. This is one of the reasons why I made _Dataset definition_ a subclass of App in the first place, because it allows this sort of generic processing.

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+          }, IMap())
+        );
+      }),
+
+      rxu.shareLatest()
+    );
+  }
+
+  get_parts(){
+    return this.parts;
```
amrc-benmorrow: Ewww Java :) . There is no advantage to this in JS unless you are going to make the fields private, and I wouldn't bother with that.

acs-data-access/lib/dataflow.js
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  get_general_infos(){
+    return this.general_infos;
+  }
+
+
+
+  async get_allowed_dataset_uuids(principal, permission, dataset_validity) {
+    const dataset_def_obs = this.get_dataset_definitions(dataset_validity);
```
amrc-benmorrow: If you convert this function to an Rx operator you can avoid the `dataset_validity` parameter everywhere.

An Rx operator is not a complicated thing, it's just something you can call from within `.pipe`; in RxJS it's implemented as a function that returns a function which modifies the source sequence. (I like JS, you can do things like that…)

So in this case you would want something like
```
filter_allowed_uuids (principal, permission) {
    const acls = this.auth.watch_acl_with_perm(principal, permission);
    return datasets => datasets.pipe(
        rx.combineLatestWith(acls),
        rx.map(([datasets, targets]) => { /* as before */ }),
    );
}
```
Then when you use it you start with the appropriate datasets sequence (all datasets or valid-only) and apply the ACL operator, i.e. still in Dataflow you define
```
metadata_list (principal) {
    return rxu.rx(
        this.valid_definitions,
        this.filter_allowed_uuids(principal, Constants.Perm.ReadDataset),
    );
}
```
where `data.valid_definitions` is defined in the constructor as
```
    this.valid_definitions = this.dataset_definitions.pipe(
        rx.filter(def => def.structure != StructurallyInvalidDataset));
```
(This assumes the change above to avoid keeping invalid definitions). Then for the API function `metadata_list` you just call
```
    const uuids = rx.firstValueFrom(this.data.metadata_list(req.auth));
```
and for the notify interface you can make a similar call, without the `firstValueFrom`, using the principal from the notify session. By pushing the `firstValueFrom` as late as possible you make the notify interface easier to implement.

acs-data-access/lib/influx-reader.js
```diff
@@ -0,0 +1,88 @@
...
+        this.influx_query_api = opts.influx_client.getQueryApi(opts.influx_org);
+    }
+
+    async get_dataset_data(topLevelInstance, meta = {}) {
+        const { from, to, measurement } = meta;
+
+        // Build range
+        const rangeClause = (from && to)
+        ? `|> range(start: ${from}, stop: ${to})`
+        : `|> range(start: -5m)`;
```
amrc-benmorrow: I don't think this is correct? I don't know Flux, really, but this looks like it's requesting the last 5 minutes. With no bounds we want all data available.

acs-data-access/lib/utils.js
```diff
@@ -0,0 +1,26 @@
+
+export function convert_to_csv (rows){
+    if (!rows.length) return "";
+
+    const headers = Object.keys(rows[0]);
+    const lines = rows.map(row =>
+        headers.map(h => JSON.stringify(row[h] ?? "")).join(",")
```
amrc-benmorrow: `JSON.stringify` is not the correct form of escaping to use here. I'm not sure what is, CSV is… vague…, but [RFC 4180](https://www.rfc-editor.org/rfc/rfc4180) is probably the best reference, and it does not use JSON string escapes.

lib/js-service-client/lib/service/data-access.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /** 
+     * Returns list of dataset uuids with READ_DATASET permission. 
+     */
+    async get_readable_dataset_uuids(){
+        const [st, json] = await this.fetch('v1/metadata');
+        if(st === 404) return [];
+        if(st == 403)
+            this.throw(`Unauthorised to access dataset ${uuid}`);
```
amrc-benmorrow: `ServiceInterface.throw` takes the status code as the second parameter. This is useful information and should be included.

amrc-benmorrow (changes requested):
Looks pretty good. I've requested a few changes. Some of them are larger-scale refactors and may not be worth pursuing at the moment.

PR #646, opened 2026-05-08
OAuth Support

AlexGodbehere:
## Summary

Adds OAuth/OIDC sign-in to ACS via a self-hosted Keycloak that federates against the F+ Auth service. Grafana now logs in through OIDC instead of being seeded with admin credentials, and Grafana roles are driven directly from F+ ACL grants. The whole stack is provisioned automatically by `service-setup` on `helm upgrade`.

## What's in the box

**Keycloak deployment**
- New `acs-keycloak` Docker image (Keycloak 26.1.1 + custom F+ User Storage SPI baked into `/opt/keycloak/providers/`).
- New `openid` Deployment, Service, Ingress and Postgres-backed JAAS config in the chart (`deploy/templates/openid/*`).
- Realm, federation provider, claim mappers and OIDC clients all created idempotently by `service-setup` (`acs-service-setup/lib/openid.js`).

**F+ User Storage SPI** (`acs-keycloak-spi/`)
- `FactoryPlusUserStorageProvider` looks users up against `acs-auth` instead of Keycloak's local user table; passwords validate against Kerberos.
- Caching store with TTL so login flows aren't HTTP-bound.
- Per-realm component config (`auth.url`, `auth.principal`, `auth.keytab.path`, cache TTL).
- SPNEGO outbound auth for the SPI's own F+ HTTP calls.
- Two protocol mappers stamp every issued token:
  - `fp_principal_uuid` - the F+ principal UUID for the signed-in user.
  - `fp_permissions` - the user's F+ permission UUIDs (multi-valued, JMESPath-friendly).
- Comprehensive Wiremock + Testcontainers integration tests.

**Grafana wiring**
- Generic OAuth flow against the new realm; admin seed account is renamed so it doesn't collide with OIDC sign-in.
- `role_attribute_path` matches `fp_permissions` against ACS-defined Grafana role groups (Admin/Editor) seeded by `service-setup`.
- `role_attribute_strict` enforced - revoking the F+ permission demotes the user on the next login. Grafana role capped at Admin (no GrafanaAdmin escalation path through ACL).

**ACS-branded login theme**
- New `factory-plus` Keycloak theme (`acs-keycloak/themes/factory-plus/`) replacing the stock `keycloak.v2` look with shadcn-vue styling that matches `acs-admin`.
- Hand-rolled CSS, FreeMarker templates for login / logout-confirm / error / info pages, and a scoped `messages_en.properties`.
- Activated via `realm.loginTheme = factory-plus` from `service-setup`; `openid.loginTheme: ""` in `values.yaml` reverts to stock.

**Supporting changes**
- `service-setup` allows per-client URL overrides (rootUrl, redirect, post-logout) and asks Keycloak to push backchannel logout to OIDC clients.
- `LocalSecret` CRD and RBAC updates so a fresh install can bootstrap Keycloak client secrets.
- `krb5.conf` gets `rdns = false` to avoid libkrb5 reverse-DNS hangs in the openid pod.
- Build infrastructure: top-level `make` is portable on macOS BSD/GNU make, parallel subdir builds via `SUBDIRS_PARALLEL`, per-subdir log files with a summary.
- `acs-mqtt`: portable maven-output filename expansion (BSD vs GNU).
- `acs-edge`: `skipLibCheck` so the top-level build doesn't trip on transitive type errors.

**Docs**
- `docs/services/oauth-clients.md` - operator guide for adding OIDC clients.
- `docs/plans/2026-05-07-keycloak-fp-spi-plan.md` - the staged plan this branch implemented (kept for archival).
- `docs/plans/2026-05-07-keycloak-fp-user-storage-spi.md` - SPI architecture notes.

## Migration

For existing deployments, `helm upgrade` is sufficient. The first run will:
1. Provision the `factory_plus` realm, the `factoryplus` federation provider, and the configured OIDC clients (`grafana` by default).
2. Mount the SPI keytab into the openid pod and seed Grafana role groups in F+.
3. Restart `openid` and `mqtt` to pick up regenerated config.

To opt out of the branded login theme set `openid.loginTheme: ""`. To run upstream Keycloak without the SPI override `openid.image.repository`/`registry` - but the F+ federation provider won't load.

## Test plan

- [x] `helm upgrade` succeeds against an existing cluster; no service-setup errors.
- [x] Realm `factory_plus` exists with `loginTheme=factory-plus`, federation provider `factoryplus`, and the `grafana` client.
- [x] Sign in to Grafana via OIDC as a user in the F+ Admin role group; user lands as Grafana Admin.
- [x] Revoke the F+ Admin grant; on next Grafana login the user is demoted (role_attribute_strict).
- [x] Sign-in page renders with the ACS branding (`+` mark + `ACS` wordmark, slate-900 button, no `&laquo;` artifact).
- [x] Logout flow renders the branded logout-confirm page; "Back to Application" link returns to the originating client.
- [x] Token introspection on a Grafana access token shows `fp_principal_uuid` and a multi-valued `fp_permissions` claim.
- [x] SPI Testcontainers integration tests pass (`mvn -pl acs-keycloak-spi verify`).
- [x] Helm upgrade (same version) or restart kerberos-keys-operators both centrally and `fplus-edge`. Validate that password for `_bootstrap` is the same before and after in `keycloak-clients`.
- [x] Follow [guide](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/blob/dn/oAuth/docs/services/oauth-clients.md#worked-example-jupyterhub) to set up 3rd party OAuth application in docs.


AlexGodbehere:
## End-to-end validation: third-party OIDC consumer (JupyterHub)

To validate that this PR isn't just "Grafana with extra steps", we ran the entire flow with **JupyterHub** as the OIDC client - installed as a separate Helm chart, not packaged with ACS, no code in this repo aware that it exists. What worked:

### What we did on the ACS side

1. **One values entry, no code change.** Added a `jupyterhub` block to `serviceSetup.config.openidClients` in the existing ACS values. `helm upgrade` and that was it: service-setup created the Keycloak client, attached the `fp_principal_uuid` and `fp_permissions` claim mappers, and dropped the auto-generated client secret into `keycloak-clients/jupyterhub`. The generic `openidClients` extension point did everything; we didn't touch service-setup logic.

2. **Created two F+ Permission objects via the ACS Admin UI.** Navigated to ConfigDB -> Create Object, selected `Permission (Rank 1 class)`, created `JupyterHub admin` and `JupyterHub user` with chosen UUIDs. These live in F+ as ordinary Permission objects - JupyterHub-specific only by name.

3. **Granted the admin permission to `admin@<realm>`** via Access Control -> Principals, with target = Wildcard. Verified the grant via the ACS MCP server (`find_grants`, `check_acl`).

### What we did on the JupyterHub side

Standard [Zero-to-JupyterHub](https://hub.jupyter.org/helm-chart/) install with `GenericOAuthenticator`, pointing the auth/token/userinfo/logout URLs at our Keycloak. The two ACS-aware bits in JupyterHub's config:

```yaml
auth_state_groups_key: oauth_user.fp_permissions
admin_groups:
  - "<jupyterhub-admin permission UUID>"
allowed_groups:
  - "<jupyterhub-admin permission UUID>"
  - "<jupyterhub-user permission UUID>"
```

`auth_state_groups_key` tells the OAuthenticator (>=17) to look in the post-OAuth `auth_state` dict for the groups list at the dotted path `oauth_user.fp_permissions` - i.e. the array of UUIDs that this PR's `acs-keycloak-spi` `FactoryPlusPermissionsMapper` stamps into every issued token. `admin_groups` and `allowed_groups` are just lists of UUIDs JH compares against that array. Membership of the admin UUID -> hub admin; membership of the user UUID -> regular access; neither -> 403.

That's the entire integration. No glue code, no shim. The application reads UUIDs out of a JWT claim and gates on UUIDs it was told about.

### Verified end-to-end

- Sign in to `http://jupyterhub.<acs.baseUrl>` -> redirected to ACS-branded Keycloak login.
- Authenticate as `admin@<realm>` (the F+ principal, password validated via Kerberos through the F+ User Storage SPI in `acs-keycloak-spi`).
- Token issued with `fp_permissions` containing the JupyterHub admin UUID we created.
- JupyterHub honours `admin_groups`, lands the user in JupyterLab as a hub admin.
- Confirmed permissions changes propagate: removed/regranted, verified the change via the userinfo endpoint, hit JH again.

### One sharp edge we found and documented

Keycloak's user attribute cache memoises `fp_permissions` per-user-load, on top of the SPI's own 60s lookup cache. Granting a new permission in F+ does *not* show up in tokens until Keycloak's cache is invalidated. Today the fix is `kubectl rollout restart deploy/openid`. Better fix queued in the docs: set `cachePolicy: NO_CACHE` on the F+ federation component so attribute changes propagate without a restart.

### Docs added in this PR

`docs/services/oauth-clients.md` now contains:
- **"Worked example: JupyterHub"** - the runbook above, complete with values, the ACS Admin UI clickthrough for creating Permissions and grants, and the Z2JH values.
- **Troubleshooting -> "Inspecting what a user's JWT actually contains"** - the technique we used to debug, including how to (temporarily) enable direct access grants on a client, password-grant a token, and decode it. Plus the cache-flush step for when grants don't propagate.
- **Constraints and gotchas** updated to note the two-layer cache (the previous "60s lag" claim under-described it).

This makes the PR's value proposition concrete: **operators can connect any OIDC application to ACS by adding one values entry and creating F+ permission objects in the UI - no ACS internals touched**. JupyterHub is the proof; Grafana is the in-tree reference.

AlexGodbehere:
## Follow-up: dropped the cache-invalidation footgun

The JupyterHub validation surfaced a sharp edge: F+ permission grants didn't propagate to consumers' tokens without a `kubectl rollout restart deploy/openid`. Two cache layers were responsible:

1. **The SPI's own 60s TTL** on F+ lookups (intentional, configurable, fine).
2. **Keycloak's per-user attribute cache** on the `UserStorageProvider` component, which defaults to `DEFAULT` (cache forever, invalidated only on restart). Once a user had been loaded into Keycloak's cache, attribute reads served the snapshot taken at first load - so a new ACL grant in F+ stayed invisible in `fp_permissions` indefinitely.

**Fix (commit 367197da):** `acs-service-setup/lib/openid.js` now sets `cachePolicy: NO_CACHE` on the `factoryplus` federation component when it provisions or updates it. Each token issuance re-reads attributes via the SPI; the SPI's `cache.ttl.seconds=60` still shields F+ Auth from per-request load. After this lands:

- Grant a permission in ACS Admin -> wait up to 60s for the SPI cache -> user signs in again -> new permission is in their `fp_permissions` claim.
- No openid restart needed for routine grant changes.

Docs updated in the same commit. `docs/services/oauth-clients.md`:

- Constraints and gotchas now describes a single 60s lag, not two cache layers.
- The JupyterHub example's "If you grant or change a permission later" section drops the openid-restart step and just says wait + re-login. Restart is mentioned only as a debugging shortcut to bypass the SPI's 60s cache.

End-state of the operator story is now: one values entry adds the OIDC client, F+ permission objects live in ConfigDB, ACL grants in Access Control, and changes propagate within a minute - all via the supported `openidClients` extension point with no ACS code changes.

AlexGodbehere:
## Follow-up: SPI bug that NO_CACHE exposed (commit `5de7c16b`)

The `cachePolicy: NO_CACHE` change in the previous commit revealed a pre-existing bug in `FactoryPlusUserAdapter`. Symptom on the test cluster (`fpd-fry`, fresh install of `5.2.0-ago.14`): Grafana sign-in fails with **HTTP 500 -> Error getting email address -> /userinfo/emails 404** (the GitHub-shape `/emails` fallback Grafana does when it can't extract an email from the userinfo response).

### What was actually wrong

Keycloak exposes user properties two ways and the SPI was handling them inconsistently:

- `getAttributes()` (bulk) - already delegated to `super` (the author fixed this previously, see the comment in the file).
- `getAttributeStream(name)` / `getFirstAttribute(name)` (per-name) - returned `Stream.empty()` / `null` for anything but the F+ attributes, masking username and email.

The standard `oidc-usermodel-attribute-mapper` Keycloak attaches to the realm's `profile` / `email` client scopes - the ones that produce the `preferred_username` and `email` claims - reads via the per-name path. With both null, tokens came out with `preferred_username: null` and `email: null`:

```json
// fry, ago.14, NO_CACHE on the federation:
{
  "sub": "f:...:...",
  "fp_principal_uuid": "...",
  "fp_permissions": [...],
  "email_verified": false
  // no preferred_username, no email
}
```

### Why we hadn't noticed

Keycloak's `UserStorageProvider` user cache (`cachePolicy: DEFAULT`) snapshots a user's attributes via `getAttributes()` on first load and serves *all* subsequent lookups from that snapshot. The bulk path was correct, so the snapshot contained username and email. Every per-name lookup hit the cache instead of our adapter, and the bug stayed hidden.

The previous commit set `cachePolicy: NO_CACHE` on the federation provider so ACL changes propagate without an openid restart. That switch sent the per-name lookups straight to the buggy SPI methods, which is what fry was hitting.

`ago` still has cache=DEFAULT so it shows the cached behaviour (`preferred_username` present, looking healthy) - but it has the same latent bug; flipping its cache off would have produced the same Grafana failure.

### The fix

`FactoryPlusUserAdapter.getAttributeStream(name)` / `getFirstAttribute(name)` now route `name=username` -> `getUsername()` and `name=email` -> `getEmail()`, and delegate everything else to `super`. Tests added that would have caught this. No change to the bulk method - it was already correct.

Net result:

- Keep `cachePolicy: NO_CACHE` from the previous commit - ACL grant changes still propagate without restart.
- Grafana's OAuth flow works again - tokens come back with `preferred_username` and `email` populated.
- The same fix applies on ago too; it just wasn't symptomatic there because of the cache.

### Why the SPI test suite didn't catch this earlier

The existing tests only asserted on `getUsername()` and `getEmail()` directly. Those still worked - what was broken was the per-name attribute methods that Keycloak's mappers actually use. New tests in `FactoryPlusUserAdapterTest` cover both paths (username, email, and the F+ attribute names).

docs/plans/2026-05-07-keycloak-fp-user-storage-spi.md
KavanPrice, line 1: Is there value in these plans being public?

deploy/templates/openid/openid.yaml
```diff
@@ -0,0 +1,170 @@
...
+          volumeMounts:
+            - mountPath: /config/krb5-conf
+              name: krb5-conf
+            - mountPath: /keytabs
+              name: krb5-keytabs-dbinit
+            - mountPath: /sql
+              name: db-init-sql
+
+      containers:
+        - name: keycloak
```
KavanPrice, line 85: Since we're pulling Keycloak in as an external service, I think it's worth putting resource limits on so it doesn't take over clusters where ACS is deployed.

KavanPrice (dismissed):

deploy/templates/openid/openid.yaml
AlexGodbehere: Yes, good idea. Feel free to add.

docs/plans/2026-05-07-keycloak-fp-user-storage-spi.md
AlexGodbehere: Probably not, although nice for traceability I guess. No strong feelings either way.

AlexGodbehere: Actually, I do like them being in here. Happy to keep or would you prefer to remove?

AlexGodbehere:
## Follow-up: drop LocalSecret CRD dependency on the central chart

Hit a fresh bug running the 5.1.0 to OAuth migration test on a freshly-nuked AGO cluster. `helm upgrade` failed with:

```
Error: UPGRADE FAILED: [resource mapping not found for name: "keycloak-admin-bootstrap" namespace: "factory-plus" from "": no matches for kind "LocalSecret" in version "factoryplus.app.amrc.co.uk/v1"
ensure CRDs are installed first, ... (same for keycloak-admin-admin, keycloak-client-grafana, keycloak-client-jupyterhub)]
```

### Root cause

The OAuth templates introduced four `LocalSecret` resources in `openid/local-secrets.yaml` so that krb-keys-operator would generate random passwords for Keycloak's admin / admin-cli / per-client credentials.

That works on a fresh `helm install` of any v5.2.x because the chart ships the `LocalSecret` CRD via `deploy/crds/` (added in 14c13caf), and Helm 3 applies entries from `crds/` on first install. **It does NOT apply them on `helm upgrade`** - that's a deliberate Helm 3 design choice to stop charts breaking existing CRD instances.

So on the upgrade path from 5.1.x (which never created any LocalSecret resources on the central chart, so never had the CRD applied) to the OAuth release, the API server has no `LocalSecret` kind and the upgrade fails before any resources are written.

The reason this never bit anyone before: the **edge-cluster** subchart ships the LocalSecret CRD in its own `crds/` directory, so any cluster that ever deployed an edge cluster has the CRD on the API server as a side-effect. Most real clusters have done that at least once. Fresh central-only clusters - exactly what nuking AGO produced - had never had the CRD applied.

### Solutions considered

1. **Pre-install/pre-upgrade Job that `kubectl apply`s the CRD** - works in every scenario (zero manual intervention, idempotent regardless of CRD origin) but costs ~60 lines of YAML (Job + ServiceAccount + ClusterRole + ClusterRoleBinding + hook annotations + the inline CRD apply). Adds a moving part on every helm operation.
2. **Move CRD to `templates/` with `helm.sh/resource-policy: keep`** - clean for clusters that get the CRD from this chart only, but breaks for clusters where edge-cluster already installed the CRD (Helm errors "resource already exists, not managed by Helm"). Hits the take-ownership wart.
3. **Drop LocalSecret on the central chart entirely** - the four LocalSecrets only exist to generate random passwords; that's the exact use case the `lookup` + `randAlphaNum` pattern already serves (we ship that today for `grafana-admin-user`).

Went with option 3. The LocalSecret pattern stays appropriate for edge agent / OPC server / driver passwords on edge clusters, where the operator's password rotation and Sealed-Secret integration matter. On the central cluster it was buying us nothing that a native Secret with `lookup` doesn't already do for free.

### What this commit does

- Replaces `deploy/templates/openid/local-secrets.yaml` with `deploy/templates/openid/keycloak-clients.yaml` - a native K8s Secret with the same name and key structure (`_bootstrap`, `_admin`, plus one key per enabled non-builtin OIDC client).
- Uses `lookup` per-key so existing values are preserved across upgrades; new keys (e.g. adding a new OIDC client) get a fresh `randAlphaNum 32` on the first upgrade that includes them.
- `helm.sh/resource-policy: keep` so `helm uninstall` doesn't take the Secret out from under any consumers.
- Removes `deploy/crds/local-secret-crd.yaml` (no longer needed on the central chart; edge-cluster subchart still ships its own copy).
- Trims `localsecrets` / `localsecrets/status` verbs from the krb-keys-operator ClusterRole, since the central operator no longer reconciles any.

Downstream consumers - Keycloak StatefulSet env (`KEYCLOAK_ADMIN_PASSWORD` from `_bootstrap`), service-setup volume mount, openid pod - are all unchanged.

### Behaviour matrix

| Cluster state | `helm upgrade --install` outcome |
|---|---|
| Fresh cluster, never had ACS | Secret created with fresh random passwords. No CRD needed. |
| 5.1.0, no edge cluster ever deployed (the AGO test case) | Secret created with fresh random passwords. No CRD needed. No manual `kubectl apply`. |
| 5.1.0 + edge clusters (LocalSecret CRD present from edge-cluster subchart) | Secret created with fresh random passwords. CRD now unused on the central side but harmless. |
| Previously on a pre-fix OAuth dev release (LocalSecrets exist, Secret is operator-owned) | `helm upgrade --take-ownership` may be needed once if the existing `keycloak-clients` Secret has owner references to the now-deleted LocalSecrets. Only affects internal dev clusters. |

The first three cases are zero-intervention.

### Edge clusters unaffected

LocalSecret usage in `edge-helm-charts/charts/edge-agent` and `edge-helm-charts/charts/opcua-server` is untouched. The `edge-cluster` subchart still ships the CRD when an edge cluster is deployed, and `krb-keys-operator` running on edge clusters still has the RBAC it needs there.

docs/plans/2026-05-07-keycloak-fp-user-storage-spi.md
KavanPrice: Happy to keep

AlexGodbehere:
## Two more issues found and fixed while running the AuthProxy to OAuth migration test

Hit both of these in the same upgrade test (5.1.0 install with AuthProxy-created Grafana users, then upgrade to a dev release and try OAuth sign-in as those existing users). Each blocks the migration entirely on its own.

### 1. JMESPath leaks `' '` instead of falling through to `'Viewer'` (7b5f4bc7)

**Symptom:** users with zero F+ permission grants couldn't sign in. Grafana refused with:

```
Integration requires a valid org role assigned in idP. Assigned role: " "
```

`role_attribute_strict: true` was working as designed - the JMESPath was returning a literal single space, which isn't a valid Grafana role.

**Diagnosis:** dumped the userinfo Keycloak issues for a no-grants user via `kcadm.sh get clients/<id>/evaluate-scopes/generate-example-userinfo`:

```json
{
  "sub": "f:fbc6a05f-e2c5-4a1e-8060-ca065e251585:...",
  "email_verified": false,
  "fp_principal_uuid": "...",
  "preferred_username": "me1newuser@AMRC-FPD-AGO.SHEF.AC.UK",
  "email": "me1newuser@AMRC-FPD-AGO.SHEF.AC.UK"
}
```

The `fp_permissions` claim is **entirely absent**. Even though our `FactoryPlusPermissionsMapper` has `multivalued: true` set, Keycloak's `OIDCAttributeMapperHelper.mapAttributeValue` collapses an empty Collection to `null` before emission, so the claim is dropped from the token rather than being emitted as `[]`.

The previous JMESPath called `contains(fp_permissions[*], '<uuid>') && 'Admin' || contains(fp_permissions[*], '<uuid>') && 'Editor' || 'Viewer'`. With the claim missing, `fp_permissions[*]` resolves to `null`, and `contains(null, '<uuid>')` interacts with go-jmespath (what Grafana uses) in a way that leaks a single space through the `||` chain rather than falling through to `'Viewer'`.

Verified against python-jmespath as a reference: with no `fp_permissions` key, `contains(fp_permissions[*], ...)` throws `JMESPathTypeError`. go-jmespath is more lenient but produces the same broken end result.

**Fix:** wrap each `fp_permissions` reference in `not_null(fp_permissions, ``[]``)` so the subject of `contains()` is guaranteed to be an array.

```yaml
role_attribute_path: >-
  contains(not_null(fp_permissions, `[]`)[*], '{{ $perms.admin }}') && 'Admin'
  || contains(not_null(fp_permissions, `[]`)[*], '{{ $perms.editor }}') && 'Editor'
  || 'Viewer'
```

Verified against all five scenarios: no claim / empty array / admin uuid / editor uuid / unrelated uuid - now resolves to the expected role in every case.

### 2. `email_verified: false` blocks email-based user lookup, breaking AuthProxy migration (28f6dd6e)

**Symptom:** for any user already in the Grafana DB from the AuthProxy era (admin and the test users), OAuth sign-in failed at the post-token-issued stage with:

```
Login Failed
user already exists
```

**Diagnosis:** Grafana's OAuth user lookup, when it encounters a pre-existing row that matches by login or email, will only link the OAuth identity to that row if the IdP claims `email_verified: true` - **OR** if `oauth_allow_insecure_email_lookup` is set in `grafana.ini`. Neither was true:

- Our SPI's `FactoryPlusUserAdapter` extends `AbstractUserAdapter`, which has `isEmailVerified()` returning `false` by default. So Keycloak emits `email_verified: false`.
- We never set `oauth_allow_insecure_email_lookup` in our Grafana config.

With both off, Grafana drops into the "create new user" path, hits a unique-constraint collision on the existing row, and surfaces "user already exists". The original AuthProxy user_id and its folder ACLs become unreachable through OAuth.

The "insecure" wording is for situations where the OAuth provider doesn't actually verify emails. Our IdP is our own Keycloak federating our own F+ Auth - emails are derived from the kerberos principal name on the F+ Principal record, which the SPI has already accepted. They're kerberos-authoritative, so it's correct to mark them verified.

**Fix:** override `isEmailVerified()` in the user adapter to return `true`. One line plus a test.

```java
@Override
public boolean isEmailVerified() {
    return true;
}
```

Tests bumped from 9 to 10 on `FactoryPlusUserAdapterTest`, total SPI test count 74, all passing.

### What this unblocks

With both fixes deployed:

- Users with no F+ permission grants resolve to Grafana `Viewer` on OAuth sign-in (instead of being rejected for an invalid role).
- AuthProxy-created users from 5.1.0 are claimed by their OAuth identities on first sign-in instead of triggering "user already exists", preserving their `user_id` (and therefore folder ACLs, dashboard ownership, etc.).

The migration test is now actually testing migration - rather than failing at the protocol layer before user lookup can happen.

### What's still on the radar

- The `role_attribute_strict: true` behaviour means F+ permission grants are still required on every OAuth login for anything above Viewer. That's by design (per the existing comment in `values.yaml`), but worth flagging in the release notes - existing AuthProxy Admins/Editors who don't have a matching F+ permission grant will be quietly demoted to Viewer on first OAuth sign-in. Manual ACLs (per-folder, per-dashboard) survive because they're per-user, not per-role.

AlexGodbehere:
## Correction + final fix: `oauth_allow_insecure_email_lookup` is the real gate

My previous comment leaned on the `email_verified` SPI fix as the migration unblock. **That was wrong** in terms of mechanism. Posting the correction with the actual root cause for the record.

### What actually controls OAuth user lookup in Grafana

`pkg/services/authn/clients/oauth.go:206-210` (Grafana main):

```go
lookupParams := login.UserLookupParams{}
allowInsecureEmailLookup := c.settingsProviderSvc.KeyValue(
    "auth", "oauth_allow_insecure_email_lookup").MustBool(false)
if allowInsecureEmailLookup {
    lookupParams.Email = &userInfo.Email
}
```

That's the entire user-lookup setup for the OAuth flow. Note that:

- `lookupParams.Login` is **never** populated on the OAuth path. The login claim from the IdP is ignored for user-lookup purposes entirely.
- `lookupParams.Email` is only populated when `oauth_allow_insecure_email_lookup` is true.
- `email_verified` has no effect on this decision — it's not consulted at all by the OAuth lookup code.

When both Email and Login in `lookupParams` are nil, `lookupByOneOf` (`pkg/services/authn/authnimpl/sync/user_sync.go:711`) skips both branches and returns `ErrUserNotFound`. Grafana then falls into the `createUser` path, which collides with the existing row's email/login uniqueness check and surfaces the unhelpful `"user already exists"` error.

So **the SPI `isEmailVerified()` change (28f6dd6e) doesn't unblock the migration on its own** — it's semantically correct for our setup but doesn't touch the relevant code path.

### Actual fix: enable the flag (commit 3983525f)

```yaml
auth.generic_oauth:
  ...
  oauth_allow_insecure_email_lookup: true
```

The "insecure" wording in the setting name had me chasing the wrong theory — it's not "trust unverified emails", it's literally "perform the email-based lookup that's otherwise disabled." For our setup the email is kerberos-authoritative by construction (derived from the F+ Principal record the SPI has already accepted), so the flag is the right thing to flip.

### Migration test result

After deploying the fixed chart, both AuthProxy-era test users (created in 5.1.0 via the AuthProxy header path) successfully signed in via OAuth and **retained their per-user folder ACLs** — confirming:

- `user_id` preserved (login matched the existing row by email)
- Origin column flipped from "Synced via Auth Proxy" to "Synced via Auth Module"
- Folder Edit grants survive because they're keyed on `user_id`, which didn't change
- `role_attribute_strict: true` correctly demotes users without F+ grants to Viewer on first OAuth login (expected, by design)

### Why keep the `isEmailVerified()` SPI change anyway

`email_verified=true` is the semantically correct claim for our IdP — emails are kerberos-authoritative. Future consumers (or future Grafana versions that *do* gate behaviour on the claim) get the right answer for free. It's the right default; it just wasn't the thing fixing the lookup here.

### Summary of fixes that actually landed this branch since the original PR

| Commit | Fix |
|---|---|
| `7c760887` | Drop LocalSecret dependency for keycloak-clients Secret (CRD-upgrade trap) |
| `b6e60bfa` | Retry ghcr.io docker login on transient flakes |
| `7b5f4bc7` | Wrap fp_permissions in `not_null(..., \`[]\`)` so role JMESPath falls through to Viewer when claim absent |
| `28f6dd6e` | SPI stamps `email_verified=true` on federated users |
| `3983525f` | Set `oauth_allow_insecure_email_lookup: true` — the actual migration unblock |

The merged effect: 5.1.x clusters upgrading through this branch get OAuth that handles both never-had-an-edge-cluster CRD state, transient registry flakes during release builds, no-F+-permission OAuth users, and the AuthProxy-to-OAuth user-id migration — all with no manual operator intervention beyond `helm upgrade`.

PR #648, opened 2026-05-13
Accept Keycloak JWTs on F+ service endpoints

AlexGodbehere:
## Summary

Implements the shaped pitch "Accept Keycloak JWTs on F+ service endpoints" (5-day appetite).

Lets human users obtain a JWT from the newly-deployed Keycloak via Postman or curl and present it to any F+ service. Each service validates the JWT, sets `req.auth` to `fp_principal_uuid`, and runs its existing ACL checks. When i3x receives a JWT, the service-client forwards it verbatim on outbound calls to ConfigDB / Directory / etc., so a single user request fans out with consistent identity end-to-end.

No new endpoint on acs-auth, no token-exchange step, no per-service token caching. Opaque session tokens, Negotiate and Basic continue to work unchanged.

### lib/js-service-api
- New JWT branch in `auth_bearer` that triggers structurally (3 base64url segments with a JWT header). Opaque tokens (random-bytes base64, no dots) still take the existing path.
- New `JwtVerifier` (`lib/jwt.js`): lazy OIDC discovery via `OIDC_DISCOVERY_URL`, jose-managed JWKS fetch + cache + `kid` rotation, validates signature/issuer/exp.
- `req.auth` is set to `fp_principal_uuid`. Raw JWT goes on `req.auth_jwt` and into an AsyncLocalStorage scope around `next()` so service-client can forward it.
- 10 unit tests via `node --test` against a local fake issuer.

### lib/js-service-client
- New `jwt-context` AsyncLocalStorage exported from the package.
- `do_fetch` checks the ALS first: if a JWT is in scope, send it verbatim as `Authorization: Bearer <JWT>` and skip the per-service token dance.

### acs-service-setup
- `ensure_client` learns about `publicClient` (no secret), `deviceAuthorizationFlowEnabled`, and PKCE S256 enforcement.

### deploy
- New `acs-cli` Keycloak client provisioned by service-setup: public, PKCE S256, Authorization Code + Device Authorization, ROPC disabled. Postman/curl redirect URIs.
- New `amrc-connectivity-stack.oidc-env` helper emitting `OIDC_DISCOVERY_URL`. Included in eight F+ service deployments: auth, cluster-manager, cmdesc, configdb, directory, files, git, i3x.

### Docs
- `docs/services/cli-and-jwt.md`: end-user / operator guide for the new flow.
- `docs/plans/2026-05-13-jwt-auth-test-playbook.md`: dev-cluster verification playbook.
- Cross-link from `oauth-clients.md`.

## How to test

Quick local verification:

1. From `lib/js-service-api`: `npm install && npm test` (10 tests, ~1s).
2. From `acs-i3x`: `npm run build && npm test` (322 tests, ~4s) — confirms no consumer-side regressions.
3. `helm template deploy --set acs.baseUrl=...,identity.realm=...,acs.organisation=...` and grep for `OIDC_DISCOVERY_URL` — should appear in all eight F+ service deployments.

Full end-to-end verification on a dev cluster: follow `docs/plans/2026-05-13-jwt-auth-test-playbook.md` after cutting a release and `helm upgrade`. It walks through Postman setup, JWT inspection, ACL filtering on the JWT principal, and regression on existing auth methods.

## Release notes

- ACS HTTP APIs now accept Keycloak-issued JWTs on `Authorization: Bearer`, alongside Basic, Negotiate and the existing opaque session tokens. Users can sign in via Keycloak (Postman: Authorization Code + PKCE; curl: Device Authorization Grant) and present the resulting JWT directly to any F+ service.
- New `acs-cli` Keycloak client provisioned by service-setup as a public client with PKCE.
- When a service receives a JWT, the service-client forwards it on outbound calls to other F+ services, so ACL decisions run against the original user across the full request chain.
- No changes to existing auth paths or to opaque-token semantics.


AlexGodbehere:
# Review: Accept Keycloak JWTs on F+ service endpoints

**Branch:** `jwt-auth`
**Worktree:** `.claude/worktrees/jwt-auth`
**PR:** #648

A user with an F+ account can now obtain a JWT from the cluster's
Keycloak (Postman or curl) and present it on any F+ HTTP API
(`Authorization: Bearer <JWT>`). Services validate the JWT against
the cluster's own realm and run their existing ACL checks against the
JWT's `fp_principal_uuid`. When a service (i3x in particular) fans
out to other F+ services to fulfil a request, it forwards the same
JWT verbatim, so the entire chain executes as the original user.

Concretely: previously i3x had to call ConfigDB / Directory using its
own Kerberos service principal, which meant it either ignored the
caller's identity (security hole) or had to perform per-call ACL
queries against the auth service (fragile, duplicates auth logic).
Now ConfigDB sees the user directly and runs its own ACL.

Three judgement calls worth flagging:

1. **No centralised token-exchange endpoint.** The shaping exercise
   considered an RFC 8693 exchange on acs-auth and rejected it: F+
   session tokens are per-service in-memory, so a centrally-issued
   token wouldn't be recognised downstream anyway. Direct JWT
   acceptance on every service is simpler and the security model is
   equivalent.

2. **JWTs accepted by default on every service that picks up the
   new lib.** No per-service opt-in flag. If `OIDC_DISCOVERY_URL` is
   unset, the auth middleware silently falls back to its old
   behaviour, so the chart toggle (`openid.enabled`) effectively
   gates JWT acceptance for the whole cluster.

3. **`fp_permissions` claim is ignored for ACL decisions.** Live ACL
   queries against F+ Auth so revocations take effect on the next
   request rather than the next token refresh.

## Walkthrough

### Auth middleware (lib/js-service-api)

- `auth_bearer` first peeks at the credential shape. A JWT (three
  base64url segments, decodable JWT header) goes to the new
  `auth_jwt` method; everything else stays on the existing
  opaque-token store path. The current opaque tokens are 66 random
  bytes base64-encoded with no dots, so the structural check is
  unambiguous.
- `JwtVerifier` lazy-fetches the OIDC discovery doc the first time a
  JWT arrives. jose's `createRemoteJWKSet` handles JWKS caching, kid
  rotation, and concurrent fetches.
- On successful JWT auth: `req.auth = fp_principal_uuid` and the raw
  JWT is stashed on `req.auth_jwt`. The middleware then runs the rest
  of the chain inside an AsyncLocalStorage scope so service-client
  can read the JWT at outbound-call time without changes to
  application code.

### Outbound forwarding (lib/js-service-client)

- `jwt-context` lives in service-client because service-api already
  depends on service-client (one-way), so this keeps the dependency
  direction clean.
- `do_fetch` checks ALS first. If a JWT is in scope, it forwards
  verbatim and skips the existing token-fetch path entirely. No
  401-retry on the JWT path: a JWT either validates or it doesn't,
  and any transient issues are handled by jose internally.

### Keycloak client provisioning (acs-service-setup)

- `ensure_client` learns about `publicClient`, PKCE enforcement, and
  the device auth grant. Existing clients (grafana etc.) are
  unaffected because they don't set the new fields.
- `acs-cli` client added to `values.yaml` under
  `serviceSetup.config.openidClients`. Public client (no secret),
  PKCE S256, Authorization Code + Device Authorization enabled, ROPC
  disabled.

### Chart wiring (deploy)

- New helper `amrc-connectivity-stack.oidc-env` emits
  `OIDC_DISCOVERY_URL` pointing at the cluster's Keycloak realm.
- Included in eight F+ service deployments: auth, cluster-manager,
  cmdesc, configdb, directory, files, git, i3x. The change in each
  template is a single helper-include line right after `REALM`.
- When `openid.enabled` is false the helper emits an empty value, so
  the auth middleware silently falls back to its pre-JWT behaviour.

### Tests

- 10 unit tests in `lib/js-service-api/test/jwt.test.js` using
  `node --test` plus a local fake Keycloak (`http.createServer` +
  jose-minted keys). Covers: valid JWT, missing
  `fp_principal_uuid`, expired, wrong issuer, bad signature, opaque
  fallback both with and without OIDC configured.
- acs-i3x's 322 existing tests still pass.

### Docs

- `docs/services/cli-and-jwt.md`: how a user obtains a JWT and uses
  it (Postman Auth Code + PKCE, curl Device Auth Grant) and how the
  services validate it.
- `docs/plans/2026-05-13-jwt-auth-test-playbook.md`: dev-cluster
  verification playbook for after a release. Includes commands to
  confirm client provisioning, env-var wiring, JWT contents,
  on-behalf-of forwarding, ACL filtering on the JWT principal, and
  regression on existing auth methods.

## How to test

1. From `lib/js-service-api`: `npm install && npm test` — 10 tests,
   ~1s.
2. From `acs-i3x`: `npm run build && npm test` — 322 existing tests,
   confirms no consumer-side regressions.
3. From the worktree root:
   ```sh
   helm template deploy \
     --set acs.baseUrl=acs.example,identity.realm=EXAMPLE.COM,acs.organisation=Example \
     | grep -A1 OIDC_DISCOVERY_URL | head -40
   ```
   Confirms the env var renders into all eight F+ service deployments.
4. For end-to-end dev-cluster verification, follow
   `docs/plans/2026-05-13-jwt-auth-test-playbook.md` after a `helm
   upgrade`.

## Decisions / open questions

- **Where AsyncLocalStorage lives.** Defined in `lib/js-service-client`
  rather than `lib/js-service-api` because service-api already depends
  on service-client. The reverse would create a cycle.
- **Audience claim ignored.** Validated `iss` and `exp` but not `aud`.
  In v1 we trust any token from our realm; aud-based scoping is a
  follow-up if needed.
- **Websocket flow unchanged.** Service-client's `ws_with_auth` still
  goes through `service_token` / `/token`. i3x doesn't do downstream
  websockets so this is fine for v1; if a future service needs WS
  auth-as-user, the same ALS pattern can be applied there.
- **Package.json `file:` paths.** The pre-existing
  `file:../lib/js-service-client` in `lib/js-service-api/package.json`
  was broken (didn't resolve from inside the package directory). I
  fixed it to `file:../js-service-client` so `npm install` works on a
  fresh checkout and tests run. The Docker build apparently has its
  own composing setup so this wasn't visible before. Worth a sanity
  check on the publish-to-npm flow.
- **Manager UI auth not touched.** Manager UI continues to use its
  current path. Whether it eventually moves to JWTs is a separate
  decision.

## Files of note

**Auth pipeline**
- `lib/js-service-api/lib/auth.js` — new JWT branch in `auth_bearer`,
  ALS wrap in `auth()`.
- `lib/js-service-api/lib/jwt.js` — `JwtVerifier`, `looks_like_jwt`.
- `lib/js-service-client/lib/jwt-context.js` — shared ALS.
- `lib/js-service-client/lib/service/fetch.js` — JWT forwarding in
  `do_fetch`.

**Provisioning**
- `acs-service-setup/lib/openid.js` — public client support.
- `deploy/values.yaml` — `acs-cli` client config.
- `deploy/templates/_helpers.tpl` — `oidc-env` helper.

**Service deployments** (8 files, one-line each)
- `deploy/templates/{auth,cluster-manager,cmdesc,configdb,directory,files,git,i3x}/...`

**Tests + docs**
- `lib/js-service-api/test/jwt.test.js`
- `docs/services/cli-and-jwt.md`
- `docs/plans/2026-05-13-jwt-auth-test-playbook.md`

AlexGodbehere:
## Follow-up: dropped `acs-configdb/package-lock.json` (1d3a6dad)

Caught a deploy-time regression in this PR when v5.2.0-i3x.4 hit a dev cluster: configdb crash-looped with `ERR_MODULE_NOT_FOUND` for `jose`, and `service-setup` stalled behind it.

### Cause

`acs-configdb/package-lock.json` had `node_modules/@amrc-factoryplus/service-api` recorded in the legacy **snapshot** form (inline `"version"`, `"resolved": "file:..."`, frozen `dependencies` map), not the **symlink** form (`{ "resolved": "../lib/js-service-api", "link": true }`) that the other consumers use. The other in-tree consumers of `lib/js-service-api` (`acs-auth`, `acs-files`, `acs-cmdesc`, `acs-i3x`) all have lockfiles in symlink form, plus `acs-directory` / `acs-cluster-manager` / `acs-git` have no lockfile at all - so configdb was the odd one out.

Under `npm install --install-links` against the snapshot form, npm trusts the lockfile's frozen dependency list and doesn't re-read the lib's live `package.json`. So when this PR added `jose` to `lib/js-service-api/package.json`, configdb's docker build copied the lib's *source* (with `jwt.js` importing `jose`) but did not install `jose` itself. Boom at startup.

The same lockfile was also stale enough to omit `rx-client` entirely (configdb's `package.json` lists it, the old lockfile predates that change).

### Fix

Dropped the file. The Dockerfile uses `npm install --install-links`, not `npm ci`, so absence is acceptable. Verified locally that a fresh install pulls `jose` plus all the `@amrc-factoryplus/*` packages including `rx-client`.

### Trade-off

Loss of reproducible-build pinning for configdb. Three other services (`acs-directory`, `acs-cluster-manager`, `acs-git`) already run this way, so it's not unprecedented in the repo. CI is the tripwire if a future transitive ever breaks the start path.

### Follow-up

Regenerate `acs-configdb/package-lock.json` in symlink form to restore reproducibility. Deferred from this PR because npm 10.8.2 in this layout refuses to materialise a top-level lockfile under `--install-links` for file: deps (only writes `node_modules/.package-lock.json`); cleaning that up is its own chore.

AlexGodbehere:
## Ready to merge

Verified end-to-end against the fpd-ago dev cluster:

- `acs-cli` Keycloak client provisioned by service-setup (public, PKCE S256, Auth Code + Device Auth, ROPC off).
- `OIDC_DISCOVERY_URL` wired to all 8 F+ services via the new `oidc-env` Helm helper.
- JWT obtained from Keycloak via Postman's Authorization Code + PKCE flow against the `acs-cli` client.
- Same JWT accepted at `auth`, `configdb`, `directory`, `i3x` (and by extension every other service that picks up the new lib).
- Auth logs show `Handling Bearer auth` → `Auth succeeded for [<my-uuid>]` on each request.

### Commits added since the original walkthrough

| Commit | Why |
|---|---|
| `589d6cfa` | Browser-safe `jwt-context` so acs-admin's Vite bundle compiles. |
| `1d3a6dad` | Drop `acs-configdb/package-lock.json`. Old "snapshot" lockfile format pinned a frozen copy of `service-api`'s dependency list and stopped `jose` from being installed; configdb crash-looped with `ERR_MODULE_NOT_FOUND`. See [#648 comment](https://github.com/AMRC-FactoryPlus/amrc-connectivity-stack/pull/648#issuecomment-4443849386) for the trade-off. |
| `7e8b7e1b` | (@DanielJohnNewbould) Null-safe `req?.log` in WebAPI's auth log adapter. Without this, every Bearer-with-JWT request died with a misleading `TypeError` before validation ran. |
| `172eedee` → `d1cf8437` | Ready-to-import Postman collection covering every HTTP-facing F+ service (127 requests, OAuth pre-configured), lives at `postman/` with a README. |

### Known follow-up (not a blocker)

`GET /v2/whoami` on the auth service returns 404 when called with a JWT. It does `find_kerberos(req.auth)`, which expects a Kerberos UPN, but JWT auth sets `req.auth` to the principal UUID directly. Worked around in the Postman collection (use `/v2/whoami/uuid` instead, which returns just the UUID and doesn't need the lookup). The proper fix is a small change to `id_whoami` to handle "auth is already a UUID" - separate PR.

### Operator note

A new release (≥ v5.2.0-i3x.6) and `helm upgrade` are required before JWTs are accepted on any cluster. Before that, the auth path silently falls back to its pre-PR behaviour (Basic / Negotiate / opaque Bearer continue working unchanged).

PR #649, opened 2026-05-14
acs-directory: don't crash on bad query input

AlexGodbehere:
## Summary

The acs-directory webapi pod crashes the entire Node process when a
route handler throws an async error. Express 4 does not forward rejected
promises from async route handlers to the error middleware, so the
rejection escapes the rxjs/express boundary as an unhandled rejection
and (with Node's current default) terminates the process. The pod then
restarts and 502s through Traefik for ~20s. Any path with `:uuid` or
`:type` that's malformed, or any lookup that returns no row, bricks the
service.

Two reproductions observed on the fpd-ago dev cluster:

```
GET /v1/alert/type/configuration
  Error: invalid input syntax for type uuid: "configuration"
      at /home/node/app/node_modules/pg/lib/native/client.js:183:15
      at async DB.query (.../pg-client/lib/index.js:63:20)
      at async Model.alert_list (.../lib/queries.js:537:21)
      at async API.alert_list (.../lib/api_v1.js:256:24)

GET /v1/link/00000000-0000-0000-0000-000000000000
  TypeError: Cannot read properties of undefined (reading 'relation')
      at API.link_single (.../lib/api_v1.js:319:59)
```

Fix: wrap every async route handler in `setup_routes` via a small
`wrap()` helper that forwards promise rejections to `next()`, so the
existing WebAPI error middleware returns a 500 instead of taking the
pod down. Also return 404 from `link_single` when the UUID is unknown
to match `alert_single`'s existing behaviour, so the second
reproduction returns a sensible status code rather than a 500.

This is a pre-existing bug, not introduced by any in-flight branch.

## How to test

1. Deploy the change to a dev cluster.
2. `kubectl get pods -n factory-plus` and note the directory-webapi
   restart count.
3. Hit the service with a valid bearer token:
   `curl -H "Authorization: Bearer $TOK" https://<host>/directory/v1/alert/type/configuration`
   - Expect a 500 with a JSON-ish body, not a 502.
4. `curl -H "Authorization: Bearer $TOK" https://<host>/directory/v1/link/00000000-0000-0000-0000-000000000000`
   - Expect a 404.
5. Confirm the restart count has not increased.

## Release notes

- Fixed a crash in `acs-directory` where malformed UUID input on
  `/v1/alert/type/:type`, `/v1/link/:uuid` (and similar) would take the
  webapi pod down, causing brief 502s for all consumers until Kubernetes
  restarted it.
- `/v1/link/:uuid` now returns 404 for unknown UUIDs, matching
  `/v1/alert/:uuid`.

### DIFF
--- a/acs-directory/lib/api_v1.js
+++ b/acs-directory/lib/api_v1.js
@@ -12,6 +12,16 @@ function fqdn_split (fqdn) {
     return fqdn.match(/([^.]*)\.(.*)/).slice(1);
 }
 
+/* Express 4 does not forward rejected promises from async route
+ * handlers to the error middleware: they surface as unhandled rejections
+ * and (with the current Node default) terminate the process. Wrap every
+ * async handler so a pg error (e.g. a malformed UUID in :uuid or :type)
+ * becomes a 5xx response instead of crashing the pod. */
+function wrap (handler) {
+    return (req, res, next) => Promise.resolve(handler(req, res, next))
+        .catch(next);
+}
+
 /* Grr. Specifying security by changing URL scheme is just stupid. */
 const _tlsmap = [
     ["http", "https"],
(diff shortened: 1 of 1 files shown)


AlexGodbehere:
# Review: acs-directory no longer crashes on malformed UUID / missing row

**Branch:** `fix-directory-pg-errors`
**Worktree:** `.claude/worktrees/fix-directory-pg-errors`
**PR:** #649

The directory webapi pod was crashing whenever an async route handler
threw an error. Express 4 (the version pinned in `acs-directory`) does
not forward rejected promises from async handlers to the error
middleware - the rejection bubbles out as an unhandled rejection and
Node terminates the process. The pod then restarts; Traefik returns
502 for ~20s; the next bad request crashes it again.

Two observed paths:

- A non-UUID string in a route param typed `:type` or `:uuid` -> pg
  raises `invalid input syntax for type uuid` -> crash.
- A well-formed UUID that doesn't exist -> `link_by_uuid` returns
  `undefined` -> `link.relation` -> TypeError -> crash.

Fix is two parts:

1. A tiny `wrap()` helper in `lib/api_v1.js` that forwards rejections
   to `next()`. Every async handler is registered through this in
   `setup_routes` (via a `h(method, ...args)` shim that does the
   `bind` + `wrap` in one step). The existing WebAPI error middleware
   then turns any thrown error into a 500.
2. `link_single` now 404s when `link == null`, matching the existing
   behaviour of `alert_single`. This stops the second reproduction
   returning a 500 for what should be a 404.

`Loader` already had a try/catch around the only async work; no other
unwrapped async handlers exist in the service (verified by grepping
all `api.get/post/put/delete` registrations in `lib/` and `bin/`).

## Walkthrough

### Crash containment
- Added `wrap()` helper. Every async route handler is now registered
  via `h(method, ...args) = wrap(this[method].bind(this, ...args))`.
- No change to handler signatures or behaviour on the happy path.
- Errors that previously crashed the pod now produce a 500 via the
  existing service-api error middleware.

### Behaviour fix in link_single
- A `GET /v1/link/<unknown-uuid>` now returns 404 instead of 500.
- Matches the existing `alert_single` pattern (and the existing XXX
  comment about the timing-attack tradeoff is copied across).

### Files of note
- `acs-directory/lib/api_v1.js`: new `wrap()` helper at top of file;
  `setup_routes` rewritten to register handlers through `h(...)`;
  `link_single` gets a null check.

## How to test

1. Deploy the change to a dev cluster (e.g. fpd-ago).
2. `kubectl -n factory-plus get pods` - record the directory-webapi
   restart count.
3. `curl -H "Authorization: Bearer $TOK" https://<host>/directory/v1/alert/type/configuration`
   - Expect a 500 with a JSON-ish body, not a 502.
4. `curl -H "Authorization: Bearer $TOK" https://<host>/directory/v1/link/00000000-0000-0000-0000-000000000000`
   - Expect a 404.
5. Repeat (3) and (4) a few times. Confirm restart count is unchanged.

## Decisions / open questions

- Stayed on Express 4 rather than bumping to Express 5 (which would
  forward async errors natively). Bumping is a wider change with its
  own risk surface; the wrap() approach is one line of behavioural
  change and matches the project style.
- Did not add a `process.on("unhandledRejection")` global guard. The
  per-route wrap is more precise and the rest of the service hasn't
  needed one historically.
- Did not address the (intentional) timing-attack concern in
  `link_single` / `alert_single` - out of scope for the crash fix.

## Files of note

- `acs-directory/lib/api_v1.js` - the only file changed.

PR #651, opened 2026-05-14
acs-configdb: resolve_upn handles UUID principals (JWT callers)

AlexGodbehere:
## Summary

`resolve_upn` in `acs-configdb/lib/auth.js` hardcoded `"kerberos"` as the identity type when resolving the caller's principal. For JWT-authenticated callers, `req.auth` is a principal **UUID**, not a Kerberos UPN, so the lookup at acs-auth (`/v2/identity/kerberos/<uuid>`) returns 410 and the configdb caller (`object_create` in `api-v2.js`, plus `client_uuid` in `model.js`) then 503s the request.

This is pre-existing - the file was last touched 2025-06-24 (`05309d7d`) and the bug has been there the whole time. It only surfaces now because the `ag/acs-i3x` branch is shipping JWT auth, and any non-Kerberos caller will hit it.

Same family as the directory `acl()` and dataflow `track_targets` fixes already merged (PR #649 on `main`).

## Reproduction

Observed on `fpd-ago`, configdb image `v5.2.0-i3x.9`:

```
POST /v2/object   { "class": "<EdgeCluster-class-uuid>" }
 -> 503

configdb log:
  princ: Failed to resolve kerberos identity 3405600f-d9f9-499c-a875-d49810d57f41: 410
  http:  >>> POST /v2/object
  http:  Handling Bearer auth
  http:  Auth succeeded for [3405600f-d9f9-499c-a875-d49810d57f41]
  http:  <<< 503 undefined
```

## The fix

Replace `this.auth.resolve_identity("kerberos", upn)` with `this.auth.resolve_principal(upn)`. `resolve_principal` uses `decode_principal` to pick the identity type by shape: UUID-form strings become type `uuid` (returned unchanged), everything else falls back to `kerberos` (the pre-JWT default). The return contract is unchanged - UUID or null - so callers in `api-v2.js` and `model.js` are unaffected.

## Scope check

Grepped `acs-configdb/lib/` and `acs-configdb/bin/` for other hardcoded `"kerberos"` identity-type assumptions; this was the only one.

## How to test

1. Authenticate to configdb as a JWT caller (e.g. via the i3X shim) where `req.auth` is a UUID.
2. `POST /v2/object { "class": "<some-class-uuid>" }` with no explicit `owner`.
3. Expect a successful create (200/201) rather than 503. The configdb log should no longer contain `Failed to resolve kerberos identity <uuid>`.
4. As a regression check, the same operation from a Kerberos-authenticated client (where `req.auth` is a UPN) should still work as before.

## Release notes

- Fix `acs-configdb` 503s on object create when authenticated via JWT (caller principal arrived as a UUID, but `resolve_upn` always asked acs-auth for a Kerberos identity).

## Out of scope (noted)

`is_root(upn)` still compares the incoming principal against `this.root` (the Kerberos UPN of the root principal). A JWT caller acting as root would not match this. Deferring - not trivially correct without a UUID-form root identity to compare against, and root-via-JWT isn't the failing case here.

### DIFF
--- a/acs-configdb/lib/auth.js
+++ b/acs-configdb/lib/auth.js
@@ -29,6 +29,9 @@ export class Auth {
     async resolve_upn (upn, maproot) {
         if (this.is_root(upn))
             return maproot;
-        return this.auth.resolve_identity("kerberos", upn);
+        /* resolve_principal handles both Kerberos UPNs (pre-JWT
+         * callers) and UUID-form principals (JWT callers) via
+         * decode_principal, returning the same UUID-or-null result. */
+        return this.auth.resolve_principal(upn);
     }
 }


AlexGodbehere:
# Review: configdb 500s on class_info and config_class_list

**Branch:** `fix-configdb-class-500`
**Worktree:** `.claude/worktrees/fix-configdb-class-500`
**PR:** #650
**Base:** `main`

Two endpoints on the ConfigDB v2 API have been returning 500 Internal
Server Error on every call:

1. `GET /v2/class/:class` (`class_info`) - PostgreSQL error
   `relation "subclasses" does not exist`.
2. `GET /v2/app/:app/class/:class` (`config_list` -> `config_class_list`) -
   PostgreSQL error `invalid input syntax for type integer` because a
   UUID string was being bound into an integer column.

Both bugs are independent of the JWT auth work currently in flight on
the `jwt-auth` branch. They were introduced earlier:

- `05309d7d` (granular ConfigDB object permissions) - `class_info`
  switched from a single permission check to returning members +
  subclasses, but the second arg to `model.class_lookup` was the
  non-existent table name `"subclasses"` (correct view name is
  `all_subclass`), and the first arg was `"membership"` (the direct
  table, not the effective view `all_membership`).
- `18570d2f` (model method to query all configs for an app) - the
  rewrite introduced a local `class_id` for the `_class_id` lookup but
  still bound the *UUID string* `klass` into the SQL parameter list.

Verification used the running configdb on the `fpd-ago` cluster
(`v5.2.0-i3x.9`) with real class/app UUIDs taken from the corresponding
list endpoints. Stack traces are in the PR body.

While editing `class_info` I also added the missing `.end()` on the 404
branch. Without it the response object was returned but never flushed.

## Walkthrough

### `class_info` handler (`acs-configdb/lib/api-v2.js`)
- `model.class_lookup(klass, "membership")` becomes
  `model.class_lookup(klass, "all_membership")` - returns effective
  members, matching `GET /v2/class/:class/member`.
- `model.class_lookup(klass, "subclasses")` becomes
  `model.class_lookup(klass, "all_subclass")` - the only existing
  effective-subclass view.
- `return res.status(404);` becomes `return res.status(404).end();` so
  the response is actually flushed when the class has no members and no
  subclasses.

### `config_class_list` (`acs-configdb/lib/model.js`)
- The query `where c.app = $1 and m.class = $2` now binds `class_id`
  (integer) for `$2` instead of `klass` (UUID string). `class_id` was
  already being looked up two lines above; the bind list just hadn't
  been updated.

## How to test

1. Pick any class UUID `C` from `GET /v2/class` and any app UUID `A`
   from `GET /v2/app`.
2. `GET /v2/class/$C` returns 200 with body
   `{ members: [...], subclasses: [...] }`.
3. `GET /v2/app/$A/class/$C` returns 200 with an array of object UUIDs.
4. Cross-check:
   - `members` matches `GET /v2/class/$C/member`.
   - `subclasses` matches `GET /v2/class/$C/subclass`.
   - Step 3's array is a subset of both `GET /v2/app/$A/object` and
     `GET /v2/class/$C/member`.

## Decisions / open questions

- Used the effective views (`all_membership`, `all_subclass`) rather
  than the direct tables. This matches the documented behaviour of the
  endpoint and the names of the sibling per-relation endpoints. If the
  intent was actually to return only direct members + direct subclasses
  that should have been wired up via the `Relations` table the same way
  every other class endpoint is.
- The redundant `class_info` handler effectively duplicates two of the
  per-relation endpoints. Not in scope to consolidate here, but worth
  flagging.

## Files of note

- `acs-configdb/lib/api-v2.js` - `class_info`
- `acs-configdb/lib/model.js` - `config_class_list`

PR #655, opened 2026-05-19
acs-keycloak-spi: refresh Kerberos credentials on expiry; delegate to lib

AlexGodbehere:
## Summary

OAuth sign-in via the Keycloak SPI breaks ~24 hours after the openid pod starts. Two clusters (AGO + the other one) went down over the weekend with the same `Cannot read from System.in` chain. Root cause: the SPI's `JaasKerberosAuthenticator` logged in via JAAS once at construction and never refreshed the resulting `Subject`. When the TGT aged out, JGSS (with `-Djavax.security.auth.useSubjectCredsOnly=false` set on the pod for the pgjdbc path) fell back to a default JAAS login, found no matching entry in `/etc/jaas/login.conf`, and landed in `Krb5LoginModule.promptForName()` — which then read `System.in`. Only a pod restart could recover.

This PR is **two commits**:

1. **`refresh Kerberos credentials on expiry`** — a self-contained fix that copies the credential-lifecycle pattern (cache `GSSCredential`, check `getRemainingLifetime()`, drop and re-login when low, one-shot retry on `GSSException`) directly into the SPI. Safe to land alone.
2. **`delegate Kerberos to lib/java-service-client`** — the proper Phase 4 wiring. Retires the duplicate by delegating to `FPGssClientKeytab` (which already implements the same renewal pattern in `FPGssPrincipal._withCreds`). The lib API is RxJava, the SPI surface is sync; we bridge with `Single.blockingGet()`.

The two are split so reviewers can verify the fix logic against the prior code before evaluating the delegation. Land them together.

## Packaging change

The second commit adds `maven-shade-plugin` to `acs-keycloak-spi/pom.xml`. The lib + its transitive runtime deps (RxJava, vavr, org.json, jakarta.json, commons-lang3, httpclient5, jetty-*) now travel in the SPI provider jar — these aren't on Keycloak's classpath. slf4j-api and Jackson stay `provided`. Final jar: ~9MB (was ~40KB).

No package relocation. Keycloak doesn't use RxJava/vavr/org.json internally, and provider classloaders are isolated from Keycloak's core anyway. Relocation would only bloat the jar and obscure stack traces.

## Release notes

`acs-keycloak-spi` &mdash; OAuth sign-in no longer fails after the openid pod's Kerberos ticket expires. Previously the pod needed a manual restart every ~24 hours; now the SPI transparently refreshes its credential.

## How to test

1. Bring up an ACS cluster on this branch. The custom openid image is built from `acs-keycloak/` with the SPI jar baked in.
2. Wait until the cached TGT has expired (default krbtgt lifetime, typically 24h). If you don't want to wait: shorten `ticket_lifetime` in the cluster `krb5.conf` ConfigMap, or revoke + re-issue the `sv1openid` principal in the KDC.
3. Attempt an OAuth login (Postman, the CLI, any device-code or browser flow that triggers `FPAuthBackedUserStore.findByUsername` &mdash; i.e. anything that uses a password/username rather than SPNEGO directly).
4. The login should succeed without a pod restart. The Keycloak log should show no `Cannot read from System.in` and no `SPNEGO token generation failed`.
5. Concurrent-load smoke: hit several logins in parallel as the credential is mid-refresh, confirm no AssertionError / classloader-init failures from the shaded lib.

## What it does NOT change

- `-Djavax.security.auth.useSubjectCredsOnly=false` stays on the pod (pgjdbc still needs it; the new SPI passes its credential to JGSS explicitly so the flag is irrelevant for the SPI path).
- `/etc/jaas/login.conf` is unchanged. The lib's inline `KrbConfiguration` is authoritative for SPI logins.
- The KerberosAuthenticator interface and its callers are unchanged. Only the implementation moved.

## Notes for reviewers

- `docs/kerberos-tgt-renewal-explainer.html` is a teaching artifact for the team explaining the Kerberos behaviour and the trap. It's tied to the first commit's diff for pedagogical reasons; the second commit shrinks the actual implementation to a thin RxJava bridge but the lesson is the same.
- `FPGssProvider(null)` is intentional. The constructor parameter is only read by the `verify*()` server-side paths, which the SPI never invokes. A future change that does would NPE loudly rather than silently misbehave.

### DIFF
--- a/acs-keycloak-spi/pom.xml
+++ b/acs-keycloak-spi/pom.xml
@@ -74,10 +74,12 @@
         <!-- F+ service client. Pin to the in-tree publication version
              (0.0.0-intree); the docker build installs lib/java-service-client
              into Maven's local cache so this resolves without needing
-             a pre-staged lib/mvn directory. The current SPI doesn't
-             import this jar yet (Phase 2 uses bare java.net.http);
-             keeping the dep declared so Phase 4's FPGssClientKeytab
-             work has a one-line wiring path. -->
+             a pre-staged lib/mvn directory.
+             JaasKerberosAuthenticator delegates Kerberos/GSS to this
+             jar's FPGssClientKeytab; the shade plugin (below) bundles
+             this jar plus its transitive runtime deps (RxJava, vavr,
+             org.json, jakarta.json, httpclient5, jetty-client, ...) so
+             they are visible to Keycloak's provider classloader. -->
         <dependency>
             <groupId>io.github.amrc-factoryplus</groupId>
             <artifactId>java-service-client</artifactId>
@@ -166,6 +168,61 @@
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.2.5</version>
+            </plugin>
+            <plugin>
+                <!-- Bundle java-service-client and its runtime deps into
+                     the SPI jar so Keycloak's provider classloader can
+                     resolve them. Keycloak ships slf4j-api, so we leave
+                     that alone; everything else (RxJava, vavr, org.json,
+                     jakarta.json, commons-lang3, httpclient5, jetty-*) is
+                     not on Keycloak's classpath and must travel with us.
+
+                     We do NOT relocate (rename) packages. Keycloak does
+                     not use RxJava or vavr internally; org.json is not
+                     on Keycloak 26's classpath either (Keycloak uses
+                     Jackson). The provider classloader is isolated from
+                     Keycloak's core, so even were there a collision the
+                     SPI would see its own copies. Relocation would only
+                     bloat the jar and obscure stack traces. -->
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.5.3</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals><goal>shade</goal></goals>
+                        <configuration>
+                            <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <shadedArtifactAttached>false</shadedArtifactAttached>
+                            <filters>
+                                <filter>
+                                    <artifact>*:*</artifact>
+                                    <excludes>
+                                        <!-- Drop signed-jar metadata that
+                                             would otherwise make the JVM
+                                             reject the shaded uber-jar. -->
+                                        <exclude>META-INF/*.SF</exclude>
+                                        <exclude>META-INF/*.DSA</exclude>
+                                        <exclude>META-INF/*.RSA</exclude>
+                                        <exclude>module-info.class</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <transformers>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                            </transformers>
+                            <artifactSet>
+                                <excludes>
+                                    <!-- Provided by Keycloak at runtime;
+                                         no need (and risky) to bundle. -->
+                                    <exclude>org.slf4j:*</exclude>
+                                    <exclude>com.fasterxml.jackson.core:*</exclude>
+                                    <exclude>org.keycloak:*</exclude>
+                                </excludes>
+                            </artifactSet>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
             <plugin>
                 <!-- Failsafe runs *IT.java in the verify phase, AFTER
--- a/acs-keycloak-spi/src/main/java/uk/co/amrc/app/factoryplus/keycloak/JaasKerberosAuthenticator.java
+++ b/acs-keycloak-spi/src/main/java/uk/co/amrc/app/factoryplus/keycloak/JaasKerberosAuthenticator.java
@@ -1,17 +1,23 @@
 /* ACS Keycloak SPI
- * Production KerberosAuthenticator: JAAS Krb5LoginModule + JGSS.
+ * Production KerberosAuthenticator: thin sync wrapper around the
+ * Factory+ Java service client's FPGssClientKeytab.
  *
- * Logs in via JAAS with the configured principal and keytab to obtain a
- * TGT, then caches a GSSCredential built from the resulting Subject.
- * Before every SPNEGO token request the cached credential's remaining
- * lifetime is checked; when it is close to expiry (or any GSSException
- * is raised during context initiation) the credential and Subject are
- * dropped and a fresh JAAS login is performed. This is the same shape
- * lib/java-service-client uses (see FPGssPrincipal).
+ * lib/java-service-client already implements the keytab login,
+ * GSSCredential caching, lifetime-aware refresh and retry-on-fault
+ * pattern (see FPGssPrincipal._withCreds). We delegate to it rather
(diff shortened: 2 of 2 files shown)


AlexGodbehere:
# Review: Keycloak SPI Kerberos TGT renewal &mdash; the two-clusters-down fix

**Branch:** `fix-keycloak-tgt-renewal`
**Worktree:** `.claude/worktrees/fix-keycloak-tgt-renewal`
**PR:** #655

The Keycloak SPI's outbound Kerberos client now refreshes its credential when the underlying TGT is about to expire (or has failed). Previously the SPI logged in via JAAS once at SPI construction, cached the resulting `Subject`, and kept reusing it forever. After ~24h the cached TGT was expired; the next user-lookup attempt failed inside JGSS with the misleading `Cannot read from System.in` because `-Djavax.security.auth.useSubjectCredsOnly=false` is set on the openid pod (for the pgjdbc Postgres login path) and the JVM&rsquo;s fallback JAAS lookup found no matching entry in `/etc/jaas/login.conf`. Only a pod restart could recover.

Two-clusters-at-once is the diagnostic tell: both clusters had their last password-based login on Friday, were quiet over the weekend (no traffic hitting the user-lookup path), then their first Monday-morning sign-in attempt failed at the same point in the failure chain.

The PR is split into two commits so the fix lands together but is reviewable in isolation:

1. **`refresh Kerberos credentials on expiry`** rewrites `JaasKerberosAuthenticator` to cache a `GSSCredential` (not just a Subject), check `getRemainingLifetime()` before every SPNEGO call, drop and re-login when low, and wrap `spnegoTokenFor()` in a one-shot retry on any `GSSException`. Passes the credential explicitly to `GSSManager.createContext(...)` so JGSS never has to look credentials up itself.

2. **`delegate Kerberos to lib/java-service-client`** retires the duplicate by delegating to the lib&rsquo;s `FPGssClientKeytab`. The lib already implements the renewal pattern (`FPGssPrincipal._withCreds`); the SPI now uses it via a thin RxJava-to-sync bridge. The pom gains `maven-shade-plugin` to bundle the lib + its transitive deps into the provider jar (Keycloak doesn&rsquo;t ship RxJava / vavr / org.json).

## Walkthrough

### Why the failure looked like a stdin prompt

- JGSS with `useSubjectCredsOnly=false` falls back to a default JAAS login when the cached Subject can&rsquo;t produce credentials.
- The fallback looks up a JAAS entry under one of JGSS&rsquo; magic names (e.g. `com.sun.security.jgss.krb5.initiate`).
- The mounted `/etc/jaas/login.conf` only defines `pgjdbc` (used by the JDBC driver), so JGSS finds no matching entry.
- With no entry, `Krb5LoginModule` runs with empty options, decides it must prompt for the principal, calls `promptForName(System.in)` and fails because pods have no stdin.
- Net effect: an expired ticket reports as &ldquo;Cannot read from System.in&rdquo;. This is the trap the original code walked into.

### Fix shape (commit 1)

- Subject and `GSSCredential` are cached together. `ensureCreds()` is the single funnel for getting a usable credential: it returns the cached one if the remaining lifetime is healthy (60s headroom), otherwise drops everything and runs a fresh `LoginContext.login()` against the keytab.
- `spnegoTokenFor()` wraps `doSpnego()` in a one-shot retry: any `GSSException` from `initSecContext` triggers `invalidate()` + fresh login + one retry, before bubbling up.
- Credential is passed explicitly to `GSSManager.createContext(name, mech, cred, lifetime)`. JGSS uses our cached credential directly. The pod&rsquo;s JVM-level flag becomes irrelevant for this code path.

### Delegation (commit 2)

- `JaasKerberosAuthenticator` is now ~80 lines: instantiate `FPGssProvider(null)` and `clientWithKeytab(...)` at construction, bridge the lib&rsquo;s `Single<GSSContext>` to sync via `.blockingGet()` in `spnegoTokenFor()`.
- `FPGssProvider`&rsquo;s constructor parameter (`FPServiceClient`) is only read by `verify*()` server-side paths we don&rsquo;t invoke from the SPI. Passing `null` is intentional; a future change that needed verify would NPE loudly.
- The credential lifecycle (cache + lifetime check + retry on `GSSException`) now lives entirely in `lib/java-service-client/.../FPGssPrincipal.java`, shared with every JVM service.
- The constructor primes the credential with a throwaway `createContextHB("HTTP@localhost")`. A `LoginException` there is a real bootstrap failure; a `GSSException` (no service ticket for HTTP/localhost) is expected and swallowed &mdash; the TGT will have been minted as a side-effect.

### Packaging

- New `maven-shade-plugin` execution in `acs-keycloak-spi/pom.xml`.
- Bundles: java-service-client + RxJava + vavr + org.json + jakarta.json + commons-lang3 + httpclient5 + jetty-client + ws.
- Excludes: slf4j-api (Keycloak ships it), Jackson (Keycloak ships it), Keycloak&rsquo;s own jars.
- No package relocation. Keycloak doesn&rsquo;t use any of the bundled libs and provider classloaders are isolated. Relocation would only bloat the jar and obscure stack traces.
- Stripped `META-INF/*.SF`, `*.DSA`, `*.RSA` and `module-info.class` so the JVM doesn&rsquo;t reject the uber-jar over signed-jar metadata or module duplication.
- Final shaded jar: ~9MB (was ~40KB).

## How to test

1. Build the openid image off this branch and roll it out to a non-prod cluster.
2. Force the TGT lifetime down to something testable: edit the cluster&rsquo;s `krb5.conf` ConfigMap, set `ticket_lifetime = 5m`, restart the openid pod.
3. After ~5 minutes (well past the new lifetime + the 60s headroom), attempt an OAuth password login &mdash; e.g. Postman against `acs-cli`, or the CLI&rsquo;s device-code flow with a `username/password` (not SPNEGO).
4. Expected: login succeeds. The Keycloak log shows no `Cannot read from System.in` and no `SPNEGO token generation failed`. There may be a brief `creds == null` debug log from the lib indicating a refresh happened.
5. Repeat the login. Subsequent ones go through the cached credential until the next refresh window.

## Decisions / open questions

- Kept the duplicate-pattern commit even though the second commit supersedes it, on the theory that &ldquo;two clusters are down right now&rdquo; warrants a reviewable bisect target should anything go wrong with the lib wiring. Easy to squash on merge if you&rsquo;d rather.
- Did not remove `-Djavax.security.auth.useSubjectCredsOnly=false`. The pgjdbc path still needs it and the SPI no longer cares about it. Could remove in a follow-up if we want to simplify.
- Did not migrate `FPAuthBackedUserStore` to `FPHttpClient` in this PR &mdash; same Phase 4 territory but unrelated to the credential bug. Worth doing once we&rsquo;re committed to the lib in the SPI; happy to do it as a follow-up.

## Files of note

- `acs-keycloak-spi/src/main/java/uk/co/amrc/app/factoryplus/keycloak/JaasKerberosAuthenticator.java` &mdash; the rewrite (commit 1 self-contained, commit 2 delegates to the lib).
- `acs-keycloak-spi/pom.xml` &mdash; new `maven-shade-plugin` configuration.
- `lib/java-service-client/.../FPGssPrincipal.java` (unchanged, for reference) &mdash; the renewal pattern we now share.
- `acs-keycloak/Dockerfile` (unchanged) &mdash; still picks up the SPI jar from `target/`.

## Immediate remediation on already-affected clusters

Until this PR lands, the unblock is one command per cluster:

```
kubectl -n factory-plus rollout restart deploy/openid
```

That mints a fresh TGT. The clusters will need that restart again in ~24h until the new image ships.

PR #658, opened 2026-05-21
update fetch in service-interface.js

amrc-za:
Adds optional support for streaming responses in `fetch` via a new `response_type: "stream"` flag.
When enabled, the `raw `response `body `stream and `headers `are returned without any automatic parsing, allowing callers to consume large or continuous payloads (e.g. CSV exports) efficiently.
Existing behaviour is unchanged for all other response types, preserving full backward compatibility.


djnewbould (approved):
LGTM! The comment on line 92 might need updating with the response tuple for a stream response type.

PR #662, opened 2026-06-08
Za/data access

amrc-za:
New DataAccess Service. Supports WebAPI and Notify API.


acs-data-access/bin/api.js
```diff
@@ -0,0 +1,95 @@
...
+const { env } = process;
+
+const fplus = await new RxClient({
+  env,
+}).init();
+
+const Version = "2.0.0";
+
+const debug = fplus.debug;
+debug.log("app", "Starting acs-data-service revision %s");
```
djnewbould, line 26: Does the substitution string need a variable here?

acs-data-access/lib/validate.js
```diff
@@ -0,0 +1,50 @@
...
+
+export function valid_grant (grant) {
+    if (!valid_uuid(grant.principal)) return false;
+    if (!valid_uuid(grant.permission)) return false;
+    if (!valid_uuid(grant.target)) return false;
+    return (grant.plural === true || grant.plural === false);
+}
+
+
+export function valid_datetime(datetime_str){
```
djnewbould, line 39: I'm not sure if you've used it before but moment.js is a pretty good library for date helper functions

djnewbould (approved):

PR #665, opened 2026-06-11
Small visualiser BIRTH optimisations

KavanPrice:
# Summary

The visualiser stutters badly (and on slower machines freezes for seconds) when BIRTH messages come through, particularly when an edge node rebirths with many devices. Two causes, both fixed here:

1. Every BIRTH was getting a full protobuf decode just to pull out a Schema_UUID we usually already had. BIRTH payloads carry the device's whole metric tree so this is the most expensive decode there is, and rebirths arrive in bursts. We now skip the decode when the node already has a schema; the first BIRTH from a new device still decodes as before so icons are unaffected.
2. Each newly discovered node also triggered an immediate full re-layout of the graph, so one rebirthing edge node with 50 devices meant 50 re-layouts. These are now coalesced behind a 250ms timer. Side effect: new nodes have no position until that layout runs, so the render loop skips them quietly instead of logging every frame.

### DIFF
--- a/acs-visualiser/public/index.js
+++ b/acs-visualiser/public/index.js
@@ -37,7 +37,7 @@ class FPlusVis {
         vis.run();
         for (const { mqtt } of clients) {
             mqtt.on("packet", vis.make_active.bind(vis));
-            mqtt.on("graph", vis.reset_graph.bind(vis));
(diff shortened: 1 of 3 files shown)


djnewbould (approved):
LGTM! Tested and works as described.

PR #695, opened 2026-07-27
Admin: move an edge deployment to another host

AlexGodbehere:
Lets an ACS Node (or bridge, or plain edge deployment) be moved onto a different host in its cluster from the Manager, instead of hand-editing raw ConfigDB or deleting and recreating the node.

## The problem

The host is chosen when a deployment is created and written to its `Applications/Edge deployment` config. `acs-edge-sync` turns it into a `kubernetes.io/hostname` node selector, so when the machine dies and is replaced the edge agent has nowhere to run. The Manager shows the host as read-only text.

The two routes out today are both bad. Hand-editing the ConfigDB JSON works but an operator will not attempt it, and it leaves every connection under the node carrying a stale `topology.hostname` plus host paths pointing at device nodes that may not exist on the replacement machine. Deleting and recreating the node destroys its UUID, its Sparkplug identity, and all of its devices and connections.

## What changed

**Stale hosts are marked.** Anywhere a hostname is shown that the cluster is not currently reporting in `status.hosts`, a "Not in cluster" pill appears next to it - the same pill in the sidebar and in the tables, so the signal is learned once. In the sidebar it is followed by one line naming the next action; in a table it carries that line as a tooltip. This is the discovery mechanism: after a machine is pulled from the cluster, the node page and the cluster tables show why nothing is scheduling. Nothing is marked when the cluster reports no hosts at all, so an edge-sync outage does not paint everything amber.

Host detail on the node page

Stale host in the cluster node table

**A move flow.** Step one picks the target from the hosts the cluster is actually reporting, plus Floating, as radio options showing each host's architecture and how much is already on it. The host it is already on sits above the list rather than in it. Each step has its own description and the header says how far through you are.

Choosing the target host

Step two shows what comes with it. For a node that is its connections and their host paths, editable in place, with devices called out as unaffected. For a bridge or a plain deployment it collapses to "nothing else references this host". One warning at the top names the target and says plainly that the host paths have not been checked against it, because ACS cannot inspect the filesystem of a Kubernetes node.

Applying it closes the dialog and confirms by toast, the way every other write in the app does. There is no separate success step.

Reviewing what moves

The host detail after the move

**Host paths are editable normally too.** `deployment.hostPaths` on a connection was previously written only by the v3 to v4 migration and was invisible in the Manager. It now has an editor in the connection dialog, for drivers that deploy as their own container, so it is not a thing you can only change while moving a node.

Host paths on a connection

## Decisions worth a look

- **Write order is connections first, deployment hostname last.** There is no transaction across N+1 ConfigDB patches. The hostname is the only field that schedules anything, so a failure part-way through leaves an unmoved deployment (a state the operator already understands) rather than a moved one with stale metadata. `updateEdgeAgentConfig` does a whole-document `put_config` of the deployment, so it also has to run before the hostname patch or it would clobber it. Connections that could not be updated are named in the result rather than swallowed into a generic error.
- **`topology.hostname` is kept in sync rather than removed.** Nothing in this repo reads it. It is written by the connection dialog and by `manager-devices.js`, and the connection store only binds `topology.cluster`. Removing the duplication is the better long-term answer, but that needs confirmation that nothing outside this repo consumes it. Worth raising separately.
- **Floating deletes the key**, matching `NewEdgeDeploymentDialog`, rather than storing the string `"Floating"`.
- **A deployment whose raw Helm `values` set `hostname`** will override the move, because `spec.values` is merged after the chart's own values in `deployments.js`. The review step detects that and warns instead of rewriting the operator's values blob.
- **The shared `Input` component cannot be styled from outside.** Its `<script setup>` `defineProps` overwrites the options block's props via `Object.assign`, so `class` and `className` never reach the element and it is fixed at `h-10`. `NewEdgeDeploymentDialog`'s `class="max-w-sm"` is being dropped this way today. The host paths editor uses a plain input matching the same styling rather than fixing `Input`, since fixing it would retroactively apply classes that other call sites have been passing to no effect. Worth a separate look.
- **Cluster admin is out of scope.** A dead machine may leave a `NotReady` node and a stuck pod. ACS changes desired state; draining or deleting the node stays a cluster admin job.

## How to test

1. Open a node in an edge cluster. The Host under Cluster Details now has a move button next to it.
2. Press it. Pick a different host and press Next.
3. Check the review step lists the node's connections, and that any with host paths show them as editable fields with the "not checked against the new machine" warning.
4. Press Move. Confirm the dialog closes, a single toast appears, and the Host detail on the node page shows the new host with the pill gone.
5. In the ConfigDB, check `Applications/Edge deployment` for the node has the new `hostname`, and that each connection's `Connection configuration` has a matching `topology.hostname`.
6. Move the node to Floating. Confirm the `hostname` key is removed from the deployment entry entirely rather than set to `"Floating"`.
7. Confirm the edge agent pod is rescheduled onto the new host and the node rebirths.
8. Take a node pinned to a hostname that is not in the cluster (or remove a host from the cluster). Confirm the amber marker appears on the node page and against that row in the cluster's Nodes table.
9. Repeat 1-4 for a bridge from the Bridges page, and for a plain deployment from the cluster's Deployments tab. Confirm the review step for these says nothing else references the host.
10. Open a connection whose driver deploys as its own container. Confirm the Host Paths section appears, that existing paths load, and that adding, editing and removing them survives a save and reopen.

## Release notes

Edge deployments can now be moved to a different host from the Manager. The Host detail on a node or bridge, and each row in a cluster's Nodes and Deployments tables, has a move action that reassigns the deployment to another host in the cluster and updates the connections underneath it. Hostnames that the cluster is no longer reporting are marked, so a deployment pinned to a machine that has been removed is visible rather than silently unschedulable. Host paths on a connection (hardware such as a serial port passed through from the host) can now be viewed and edited in the Manager, both while moving a node and from the connection dialog.


AlexGodbehere:
# Review: move an edge deployment to another host

**Branch:** `move-deployment-host`
**Worktree:** `.claude/worktrees/move-deployment-host`
**PR:** #695

A deployment's host was set once at creation and never editable again. When the machine behind it is replaced, the ACS Node stays pinned to a hostname that no longer exists in the cluster, its edge agent is unschedulable, and the Manager shows the host as read-only text. The only routes out were hand-editing raw ConfigDB (which leaves stale `topology.hostname` on every connection and stale host paths) or deleting and recreating the node (which destroys its UUID, Sparkplug identity, devices and connections).

The host is now moveable from the Manager, with the connections underneath it brought along. Three things ship together:

1. A "Not in cluster" pill wherever a hostname is displayed that the cluster is not reporting in `status.hosts`.
2. A two-step move flow: pick a target host, review what comes with it, apply. Success closes the dialog and confirms by toast.
3. A host paths editor, used in the move flow and also added to the connection dialog so host paths are not only editable while moving something.

Revised against the design handover for PR #695; the write path is unchanged from the first review.

Worked example. `Cell Gateway` is pinned to `fpcgw-pf550f40`, which has been pulled from the cluster. Moving it to `fpcgw-pf662a11` writes:

- `Connection configuration` for each of its 3 connections: `topology.hostname` → `fpcgw-pf662a11`, and `deployment.hostPaths` for any whose paths were edited.
- `updateEdgeAgentConfig` for each edited connection, which rebuilds `values.drivers` on the deployment doc.
- `Edge deployment` for the node: `hostname` → `fpcgw-pf662a11`, written last.

Moving to Floating writes `{ hostname: null }`, which a JSON merge patch removes, so the key is deleted rather than set to the string `"Floating"` (matching the create dialog).

## Walkthrough

### Stale host marking
- One pill, `StaleHostPill`, used in the node and bridge sidebars and in the cluster's Nodes and Deployments tables, so the signal is learned once rather than being an amber sentence in one place and a bare triangle in another.
- In a table the pill carries the explanation as a tooltip; in the sidebar, which has the room, it sits below and names the next action ("Move the node to a reported host") rather than restating the problem twice.
- Table rows are fed from a `_hostStale` flag computed on the page (same pattern as the existing `_canRebirth`).
- Nothing is marked when the cluster reports zero hosts. An edge-sync outage or an unbootstrapped cluster should not paint every row amber.

### The move flow
- Each step has its own description, and a "Step N of 2" counter with progress bars sits in the header. Previously step one's description stayed on screen during the review, misdescribing the page.
- Target picker lists the cluster's reported hosts as radio options with a visible selected state, showing architecture and current occupancy (nodes plus deployments already on that host), then Floating in its own group with its consequence spelled out. The host the deployment is already on is shown once above the list rather than sitting in it as an unpickable row, so Next is simply disabled until something is chosen.
- Review step leads with the from/to pair, then one warning naming the target host about unverified host paths, rather than repeating a warning under every connection. It lists the node's connections with their host paths as editable fields and states the device count as unaffected.
- For a bridge or plain deployment the review collapses to "nothing else references this host". Same dialog, same code path, driven by a `kind` on the event payload.
- A deployment whose raw Helm `values` set `hostname` gets a specific warning, because `spec.values` is merged after the chart's own values in `deployments.js:233` and would silently override the move.

### Confirmation
- The third "move complete" step is gone. It duplicated the toast firing alongside it, and every other write in this app confirms by toast alone.
- Success closes the dialog and lets the Host detail re-render with the new hostname and no pill. Per-connection failures are still named individually, now in the error toast.

### Write ordering
- Connections are patched first, then `updateEdgeAgentConfig` for any whose host paths changed, then the deployment's own `hostname`.
- The ordering is load-bearing twice over: the hostname is the only field that schedules anything, so a partial failure leaves an unmoved deployment rather than a moved one with stale metadata; and `updateEdgeAgentConfig` does a whole-document `put_config` of the deployment, so running it after the hostname patch would clobber it.
- Failures are collected per connection and named back to the operator, rather than collapsing into one generic error toast.

### Host paths
- `deployment.hostPaths` was previously written only by `manager-devices.js:185` during the v3 to v4 migration and had no UI at all.
- Column labels appear once per connection rather than once per row, inputs are monospaced at 32px, and the remove control is a quiet cross that reddens on hover instead of a red bin, which read as "delete the connection".
- The connection dialog section is gated on `selectedDriver.definition.image`, not `selectedDriver.image`. The image lives on the driver definition, which is how `updateEdgeAgentConfig` finds it; the existing "Split" badge in the driver dropdown reads `driver.image` and looks like it never renders.
- Blank rows are dropped on save, and an empty list writes `hostPaths: null` so the key is removed rather than left as `[]`.
- That dialog also gains a Cancel button, and its save reads "Save Connection" when editing since the title already says Edit.

## How to test

1. Open a node in an edge cluster. The Host under Cluster Details has a move button next to it.
2. Press it, pick a different host, press Next.
3. Check the review lists the node's connections, with host paths as editable fields and the "not checked against the new machine" warning.
4. Press Move. Confirm the dialog closes, one toast appears, and the Host detail shows the new host with the pill gone.
5. In the ConfigDB, check `Applications/Edge deployment` for the node has the new `hostname`, and each connection's `Connection configuration` has a matching `topology.hostname`.
6. Move the node to Floating. Confirm `hostname` is removed from the deployment entry, not set to `"Floating"`.
7. Confirm the edge agent pod is rescheduled and the node rebirths.
8. Point a node at a hostname not in the cluster. Confirm the amber marker on the node page and in the cluster's Nodes table.
9. Repeat 1 to 4 for a bridge from the Bridges page and a plain deployment from the cluster's Deployments tab.
10. Open a connection whose driver deploys as its own container. Confirm Host Paths appears, existing paths load, and add/edit/remove survives save and reopen.

## Decisions / open questions

- **`topology.hostname` is synced, not deleted.** I could not find a single consumer of it in this repo. It is written by the connection dialog and the migration; the connection store only binds `topology.cluster`. Syncing it is the safe call for this change, but if nothing outside the repo reads it (your own tooling, i3X, anything at a customer site) then deleting the field is the better fix and should be raised separately.
- **Host paths are prefilled and warned about, never verified.** ACS cannot inspect a Kubernetes node's filesystem. Having the edge agent report available device paths per host would make the picker real, but that is new agent and ConfigDB plumbing and well outside this change.
- **The shared `Input` component silently drops `class` and `className`.** Its `<script setup>` `defineProps` overwrites the options block's props via `Object.assign`, so those props are never registered and the element is stuck at `h-10`. I confirmed in the browser that `NewEdgeDeploymentDialog`'s `class="max-w-sm"` is being dropped today for the same reason. The host paths editor uses a plain input carrying the same styling at the height the design asks for. I did not fix `Input`: doing so would retroactively apply classes that existing call sites have been passing to no effect, changing layouts outside this change. Worth its own look.
- **Cluster admin is out of bounds.** A dead machine can leave a `NotReady` node and a pod stuck Terminating. This changes desired state only.
- **No pre-assignment to hosts that have not joined.** The picker offers what the cluster reports plus Floating; no free-text hostnames.
- **No host-level bulk move.** "Move everything on host A to host B" is a different feature.
- **Screenshots** were taken by rendering the real pages against a fake ConfigDB (a temporary Vite entry point, since the Manager needs a live backend). That harness is not committed. It exercised the actual write path end to end, including merge-patch key deletion, so it is close to a functional test. Say the word if it is worth keeping as one.

## Files of note

**New**
- `acs-admin/src/components/EdgeManager/Nodes/StaleHostPill.vue` - the shared "Not in cluster" pill
- `acs-admin/src/utils/moveHost.js` - the move operation and its write ordering
- `acs-admin/src/components/EdgeManager/Nodes/MoveHostDialog.vue` - the two-step flow
- `acs-admin/src/components/EdgeManager/Connections/HostPathsEditor.vue` - shared host path editor
- `acs-admin/src/utils/hosts.js` - is this hostname one the cluster reports
- `acs-admin/src/components/EdgeManager/Nodes/MoveHostButton.vue` - row-level affordance

**Changed**
- `acs-admin/src/pages/EdgeManager/Nodes/Node.vue`, `src/pages/Bridges/Bridges.vue` - Host detail gains the action and warning
- `acs-admin/src/pages/EdgeManager/EdgeClusters/{EdgeCluster.vue,nodeColumns.ts,deploymentColumns.ts}` - stale marker and per-row move
- `acs-admin/src/components/EdgeManager/Connections/NewConnectionDialog.vue` - host paths section
- `acs-admin/src/components/SidebarDetail.vue` - optional warning and action slot
- `docs/architecture/edge-management/edge-deployments.md` - walkthrough with screenshots

PR #697, opened 2026-07-27
Keycloak login theme: match the acs-admin login card

AlexGodbehere:
## Before / after

Same viewport, same realm settings. Left is the theme as it stands on `main`,
right is this branch.

| Before (`main`) | After (this branch) |
|---|---|
|  |  |

Field-error state, which is the path most likely to regress:

These were shot against a realm configured the way `acs-service-setup`'s
`ensure_realm` provisions one: `loginWithEmailAllowed: false`,
`registrationAllowed: false`, and `resetPasswordAllowed` left at Keycloak's
default of `false` (`ensure_realm` does not set it). That is why the field
reads "Username" rather than "Username or email", and why there is no
Register or Forgot Password link. **This PR does not enable
self-registration**; those links are absent because ACS realms disable them,
and the theme renders them only when the realm asks for them.

## Summary

The Keycloak login page (the ACS-branded `factory-plus` theme) resembled the
acs-admin login card but rendered noticeably smaller and less polished. This
makes it like-for-like.

The root cause of "smaller" was one line: the theme's stylesheet set
`font-size: 14px` on `html, body` while authoring every dimension in `rem`,
so the whole page rendered at 87.5% scale. acs-admin has no root font-size
override, so it runs at the browser default 16px. Removing the override and
correcting the two values that had been authored against a 14px root brings
the geometry into exact agreement.

Measured in headless Chrome against the real stylesheet:

| | admin target | theme now |
|---|---|---|
| card width (`max-w-sm`) | 384px | 384px |
| card padding (`p-6`) | 24px | 24px |
| input height (`h-10`) | 40px | 40px |
| button height (`h-10`) | 40px | 40px |
| title (`text-2xl`) | 24px | 24px |
| label (`text-sm`) | 14px | 14px |

On top of the scale fix:

* The logo/wordmark brand row is gone; the card leads with the title, as
  admin does. The favicon link is untouched.
* The card description is finally rendered. `.acs-card-description` had
  existed in the CSS with no markup ever using it, which is most of why the
  card looked bare. `registrationLayout` gained an optional `description`
  section; pages that don't supply one render no paragraph at all.
* Title copy is now "Welcome to ACS!" and the username field has an
  "e.g. me1ago" placeholder, both from `messages_en.properties` so they stay
  translatable and overridable.
* The submit button swaps its label for a spinning notched circle while the
  form posts, matching admin's `fa-circle-notch`. It is an inline SVG with a
  CSS keyframe, not Font Awesome, because the pod must not depend on a CDN.
  The existing `prefers-reduced-motion` block stops the spin but keeps the
  marker.
* Colour parity with what admin *renders*, not what its source appears to
  intend: admin's `Input` references shadcn tokens (`border-input`,
  `ring-ring`, …) that its own build never defines, so it falls back to
  Tailwind defaults. The theme's input border is now gray-200 and its focus
  ring blue-500/50 to match. A comment in the CSS records why, so nobody
  "fixes" it back into slate. The button ring is left near-black, because
  `ring-slate-950` is a real colour and admin's button ring genuinely is.
* The social-provider buttons no longer emit a Font Awesome `<i>`; the theme
  loads no icon font, so it could only ever have rendered an empty box.
  Nothing configures social providers today, so this is invisible in practice.

## How to test

The login page can be rendered locally without a cluster:

1. Start Keycloak with the theme mounted:

   ```sh
   docker run -d --name kc-theme -p 18080:8080 \
     -e KC_BOOTSTRAP_ADMIN_USERNAME=admin -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
     -v "$PWD/acs-keycloak/themes/factory-plus:/opt/keycloak/themes/factory-plus:ro" \
     quay.io/keycloak/keycloak:26.6.3 start-dev
   ```

2. Create a realm configured the way `acs-service-setup` provisions one, so
   you see what an operator sees rather than Keycloak's `master` defaults:

   ```sh
   docker exec kc-theme /opt/keycloak/bin/kcadm.sh config credentials \
     --server http://localhost:8080 --realm master --user admin --password admin
   docker exec kc-theme /opt/keycloak/bin/kcadm.sh create realms \
     -s realm=factory-plus -s enabled=true \
     -s loginWithEmailAllowed=false -s registrationAllowed=false \
     -s sslRequired=none -s loginTheme=factory-plus
   ```

3. Open the login page:

   ```
   http://127.0.0.1:18080/realms/factory-plus/protocol/openid-connect/auth?client_id=account-console&redirect_uri=http%3A%2F%2F127.0.0.1%3A18080%2Frealms%2Ffactory-plus%2Faccount%2F&response_type=code&scope=openid&state=x&nonce=y&code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM&code_challenge_method=S256
   ```

   Compare it side by side with acs-admin's login screen. To see the previous
   version alongside it, check out `main`'s copy of
   `acs-keycloak/themes/factory-plus` into a second directory, mount it as a
   second theme, and point a second realm at it.

4. Check the error state: submit a wrong username/password. The username
   field should go red with a message beneath it, and the card should not
   shift or lose its header.

5. Check the submit spinner: enter credentials and click Sign In. The label
   is replaced by a spinning notched circle and the button disables.

6. Check the pages that supply no description still render: an invalid
   `client_id` on the auth endpoint gives the "We are sorry..." error card.
   For the inherited-template case, temporarily set `resetPasswordAllowed=true`
   on the realm and click "Forgot Password?" (an ACS realm has it off, since
   `ensure_realm` never sets it and Keycloak defaults it to false). Neither
   page should show a blank line under the title.

7. Tear down with `docker rm -f kc-theme`.

In a cluster, this ships in the `acs-keycloak` image; the theme is activated
by `loginTheme=factory-plus` on the realm, which service-setup already sets.

## Release notes

The ACS-branded Keycloak login page now matches the ACS Manager login page
exactly. It previously rendered at 87.5% of its intended size, which made it
look like a smaller, rougher copy of the Manager's login card. The page also
gains the "Enter your details below to login to your account" subtitle, a
username placeholder, and a spinner on the Sign In button while logging in.


AlexGodbehere:
# Review: Keycloak login theme rendered at 87.5% scale; now like-for-like with acs-admin

**Branch:** `agent-ae4e8c4b4fbdf85e3`
**Worktree:** `.claude/worktrees/agent-ae4e8c4b4fbdf85e3`
**PR:** #697

## Before / after

| Before (`main`) | After (this branch) |
|---|---|
|  |  |

Field-error state:



Both shots are at the same viewport, from a realm configured the way
`acs-service-setup`'s `ensure_realm` provisions one: `loginWithEmailAllowed`
false, `registrationAllowed` false, `resetPasswordAllowed` left at Keycloak's
default of false. Hence "Username" rather than "Username or email", and no
Register or Forgot Password link. Nothing was edited out of the images.

The ACS-branded Keycloak login page looked like a smaller, rougher copy of
the acs-admin login card. One line explained the "smaller": the theme
stylesheet set `font-size: 14px` on `html, body`, and every dimension in
that file is authored in `rem`. acs-admin has no root font-size override, so
it runs at the browser default 16px. The theme was therefore rendering the
entire page at 87.5%.

Worked example: `.acs-input { height: 2.5rem }` was intended as admin's
`h-10` = 40px, but at a 14px root it computed to 35px. The button, the card
padding, the title and the label were all off by the same 12.5%. Two values
had been hand-corrected against the wrong root and needed reverting: the
card `max-width` (26rem, i.e. 364px at 14px root, chasing admin's 384px) and
the card padding (1.75rem = 24.5px, chasing admin's 24px). With the root
back at 16px they become 24rem and 1.5rem and land exactly.

Measured in headless Chrome against the real stylesheet: card 384px wide,
24px padding, 40px input, 40px button, 24px title, 14px label. All six
targets hit exactly.

The rest is the polish gap. The card had a logo/wordmark brand row admin
doesn't have, and never rendered a description at all, even though
`.acs-card-description` had been sitting in the CSS unused. Colours were
matched to what admin's source appears to intend rather than what it
renders.

## Walkthrough

### Scale
- Removed the `font-size: 14px` root override, with a comment explaining why
  it must not come back.
- `max-width: 26rem` to `24rem` (`max-w-sm`), padding `1.75rem` to `1.5rem`
  (`p-6`), in the header, content and footer.
- Audited the rest of the file for values that only make sense at a 14px
  root. Found none; everything else already assumed 16px, which is the tell
  that the 14px line was the mistake.

### Structure
- Dropped the brand row (logo plus "ACS" wordmark) and its CSS. The favicon
  link in `<head>` is untouched.
- Added an optional `description` section to the `registrationLayout` macro.
  It is captured into a local and only wrapped in a `<p>` when it has
  content, so pages that do not define the section (error.ftl,
  logout-confirm.ftl, info.ftl and the upstream `keycloak.v2` templates we
  inherit) emit no paragraph rather than a blank line under the title. I
  verified this rather than assuming it.
- Header gap goes to 0. This project's `CardHeader` is `flex flex-col p-6`
  with no `space-y-1.5`, so admin's title and description nearly touch. It
  reads tight, but matching admin was the explicit requirement.
- Title copy ("Welcome to ACS!"), description, and the username placeholder
  ("e.g. me1ago") all live in `messages_en.properties`, not the template.

### Colour parity
- Input border is now gray-200 (#e5e7eb) and the focus ring blue-500 at 50%.
  This is deliberate and the CSS says so at length: admin's `Input.vue` asks
  for `border-input`, `bg-background`, `ring-ring` and
  `ring-offset-background`, but its `tailwind.config.js` has `colors: {}` and
  `main.css` defines only `--radius`. Those classes are no-ops, so admin
  renders with Tailwind's fallbacks. We match the pixels, not the intent.
- Card hover ring moved from 4px to 5px: a bare Tailwind `ring` is 3px and
  `ring-offset-2` sits inside it. Still gated behind
  `@media (hover: hover) and (pointer: fine)` with the 300ms transition.
- Button focus ring left alone. `ring-slate-950` and `ring-offset-white` are
  real Tailwind colours, so admin's button ring genuinely is near-black.

### Submit feedback
- The Sign In button now hides its label and shows a spinning notched circle
  while the form posts, matching admin's `fa-circle-notch`. Inline SVG plus a
  CSS keyframe, because the pod must not depend on a CDN.
- The existing `prefers-reduced-motion` block stops the rotation but keeps
  the spinner visible as a "working" marker.
- The social-provider buttons no longer emit a Font Awesome `<i>`. The theme
  loads no icon font, so the markup could only ever have been an empty box.
  I dropped it rather than leaving a comment; nothing configures social
  providers today so this is invisible in practice.

## How to test

1. Start Keycloak with the theme mounted:

   ```sh
   docker run -d --name kc-theme -p 18080:8080 \
     -e KC_BOOTSTRAP_ADMIN_USERNAME=admin -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
     -v "$PWD/acs-keycloak/themes/factory-plus:/opt/keycloak/themes/factory-plus:ro" \
     quay.io/keycloak/keycloak:26.6.3 start-dev
   ```

2. Create a realm shaped like the ones `acs-service-setup` provisions, rather
   than using Keycloak's `master` defaults:

   ```sh
   docker exec kc-theme /opt/keycloak/bin/kcadm.sh config credentials \
     --server http://localhost:8080 --realm master --user admin --password admin
   docker exec kc-theme /opt/keycloak/bin/kcadm.sh create realms \
     -s realm=factory-plus -s enabled=true \
     -s loginWithEmailAllowed=false -s registrationAllowed=false \
     -s sslRequired=none -s loginTheme=factory-plus
   ```

3. Hit the auth endpoint for that realm (`client_id=account-console`, with
   PKCE params; the exact URL is in the PR body). You should get the themed
   login page: "Welcome to ACS!", the subtitle underneath, a 384px card.
   Compare side by side with acs-admin's login screen.
4. Submit a wrong username/password. The username field goes red with a
   message beneath it; the header does not shift.
5. Enter credentials and click Sign In. The label is replaced by a spinning
   notched circle and the button disables.
6. Hit the auth endpoint with an invalid `client_id` for the "We are sorry..."
   card. For the inherited-template case, temporarily set
   `resetPasswordAllowed=true` on the realm and click "Forgot Password?" (an
   ACS realm has it off: `ensure_realm` never sets it and Keycloak defaults
   it to false). Neither page should show a blank line under its title.
7. `docker rm -f kc-theme`.

## Decisions / open questions

- **Placeholder colour left as slate-400.** Strictly, admin's
  `placeholder:text-muted-foreground` is another undefined token, so admin
  renders the browser default placeholder colour. The brief enumerated the
  exact colour changes it wanted and this was not among them, so I left it.
  Say the word if you want full fidelity here too.
- **Title line-height 1.1 to 1.** Not in the brief, but admin's `CardTitle`
  is `leading-none`, and with the description now sitting flush the 0.1
  difference was visible. Same for the description, which now uses
  Tailwind's `text-sm` line-height of 20px rather than inheriting the body's
  21px.
- **The description is conditional, not always-rendered.** A bare
  `<#nested "description">` inside a `<p>` would leave an empty paragraph
  taking up a line on every page that doesn't define the section. Capturing
  into a local and testing `?has_content` avoids that. Confirmed working on
  a real Keycloak.
- **Upstream page interiors are still unstyled.** Pages we inherit rather
  than override (password reset, registration) render our card shell around
  raw PatternFly form markup. That is pre-existing and out of scope here,
  but it is visible if a reviewer clicks "Forgot Password?".

## What I verified, and what I did not

Verified by running Keycloak 26.6.3 in Docker with both this branch's theme
and `main`'s copy bind-mounted, and two realms configured to match
`ensure_realm` pointed at them:

- All four theme templates parse and render. No FreeMarker errors.
- The login page renders with the title, description, placeholder and
  spinner markup, screenshotted in headless Chrome at 1000x640 against the
  live server, both before and after.
- The error state renders: bad credentials produce `acs-input-error` on the
  field and an `acs-field-error` span, with the header intact. The
  screenshot is the server's own response HTML, re-loaded from disk so it
  could be captured (the error page is a POST response, so there is no URL
  to point a headless browser at); the stylesheet is still fetched live from
  the running Keycloak.
- Pages with no description section render no empty paragraph. Checked our
  own `error.ftl` (invalid `client_id`) and an inherited upstream
  `keycloak.v2` template (`reset-credentials`).
- The six pixel dimensions, measured via `getComputedStyle` and layout
  metrics in headless Chrome against the real stylesheet.
- Rebased onto `origin/main` (which had gained #695 and #696) with no
  conflicts, and force-pushed. The three logical commits are intact, plus a
  fourth for the screenshots.

Not verified:

- I did not build the `acs-keycloak` image. It needs the SPI Maven build and
  adds nothing to a theme-only change; I mounted the theme into the stock
  upstream image at the exact path the Dockerfile copies it to.
- I did not deploy to a cluster or run any helm/kubectl.
- The alert banner (`.acs-alert`) was only rendered from a hand-built probe
  page using the real stylesheet, not from a live Keycloak, because the
  login flow suppresses the banner whenever the error is attached to a
  field. It renders correctly in the probe. There is no committed screenshot
  of it, because it would not be genuine server output.
- The submit spinner is verified as markup and CSS and was rendered from a
  probe page, but there is no committed screenshot: capturing the in-flight
  state needs a real browser session mid-POST.
- No automated tests exist for this theme, and I did not run the repo-wide
  suite, which does not cover it.
- Hover and focus-visible states were reasoned from the CSS, not clicked
  through by hand, apart from the autofocused username field which shows the
  new blue ring in the screenshot.

## Files of note

Markup and copy:
- `acs-keycloak/themes/factory-plus/login/template.ftl` - brand row removed,
  optional description section added
- `acs-keycloak/themes/factory-plus/login/login.ftl` - description,
  placeholder, spinner, social icon removal
- `acs-keycloak/themes/factory-plus/login/messages/messages_en.properties` -
  title, description and placeholder strings

Styling:
- `acs-keycloak/themes/factory-plus/login/resources/css/styles.css` - the
  scale fix, header/content structure, colour parity, spinner keyframe

Evidence:
- `docs/assets/keycloak-login/login-before.png`,
  `login-after.png`, `login-error-state.png`

PR #698, opened 2026-07-27
Keycloak SPI: cross-realm Kerberos login, and fix realm casing on user lookup

AlexGodbehere:
## What this does

Lets a user from a trusted foreign Kerberos realm log in to a cluster
that has no Factory+ principal for them, without an admin hand-creating
one first. Also fixes a casing bug that broke fully qualified logins for
local users.

### The casing bug (stands alone)

Keycloak lower-cases the username before it reaches a user storage
provider, so signing in as `me1ago@AMRC-FP.SHEF.AC.UK` arrived at
Factory+ Auth as `me1ago@amrc-fp.shef.ac.uk`. F+ Auth compares identity
names with exact string equality, so the lookup 410d and Keycloak
reported `user_not_found`. Any local user typing their full UPN instead
of their short name hit this.

The lookup now tries the verbatim form first, then retries with the
realm portion upper-cased. Retry rather than unconditional upper-casing,
so a deployment whose realm genuinely is lower case keeps working and
pays no extra request.

Known limitation: only the realm is touched. An identity stored with
capitals in the *user* portion cannot be matched, because Keycloak has
already folded the case by the time we see it.

### Cross-realm login

With `openid.trustedRealms` set, a user from a listed realm can sign in
with no local principal. When every casing candidate misses locally and
the realm is explicitly trusted, the lookup returns a *provisional*
user; the provider writes their Kerberos identity via the existing
`PUT /v2/principal/{uuid}/kerberos`, and only after the KDC has accepted
the password. They appear in the ACL editor by their UPN with zero
permissions until an admin grants some.

Default config trusts nothing, so the whole path is inert until opted
into. Two modes per realm:

- **Recommended: no `authUrl`.** The principal UUID is derived from the
  UPN as a UUIDv5, so every cluster independently derives the same UUID
  for the same person. No grant to obtain, no outbound call at login, no
  dependency on the home cluster existing.
- **With an `authUrl`** the principal UUID is resolved from that realm's
  own F+ Auth, so the user keeps that cluster's identity here. Needs a
  principal created for our `sv1openid` on the home cluster and a
  `ReadKrb` grant there.

### Derived UUIDs, and why they matter

The mint path uses UUIDv5 (RFC 4122 name-based, SHA-1) over a fixed
namespace and the canonicalised UPN, rather than a random UUID. That
buys two things:

1. **Every cluster agrees**, with no coordination and no outbound calls.
2. **Admins can pre-grant.** Compute the UUID and attach permissions
   before the user's first login, so it lands with the access they need
   instead of with nothing. This is the main reason for it.

Namespace: `cb3714c4-c85c-482a-987b-408293aa141e`, fixed forever -
changing it orphans every principal ever derived from it, so it is
pinned by a test.

Name input: the canonicalised UPN, i.e. realm upper-cased, user portion
exactly as received from Keycloak (lower-cased, since Keycloak folds it).
Nothing local feeds in - that is what makes clusters agree.

To compute one by hand:

```sh
python3 -c "import uuid; print(uuid.uuid5(uuid.UUID(
  'cb3714c4-c85c-482a-987b-408293aa141e'), 'me1ago@AMRC-FP.SHEF.AC.UK'))"
# 12bb7b35-16c8-572d-af5f-c1f312ec4ae8
```

That exact pair is a known-answer test, so the code cannot drift from
what the command produces. Implemented by hand: the JDK has no v5
generator and `UUID.nameUUIDFromBytes` is MD5/v3, not a substitute.

### Pick a mode and stay: close to a one-way door

Adding an `authUrl` to a realm that has been running without one does
not migrate anyone. Users who already logged in keep their derived UUID
while first-time users get home-derived ones, splitting the estate by
when each person first signed in. Converging afterwards means
re-pointing identity records by hand.

### The `ReadKrb` grant, stated plainly

`ReadKrb` cannot be scoped to individual principals. It is Wildcard-only
and acs-auth treats it as blanket "read any identity"
(`acs-auth/lib/dataflow.js:273`). Granting it to a consuming cluster's
`sv1openid` lets that cluster's Keycloak read and enumerate **every
identity record in the home cluster's auth service**, as a standing
capability, regardless of who logs in or whether anyone ever does. It is
also the revocation point for the whole trust relationship.

The no-`authUrl` mode is now the documented recommendation, not a
fallback. The cost is only that you do not reuse the home cluster's
particular UUID; the derived one is consistent everywhere else.

If the grant is missing the home cluster returns 403 and the login
fails. It deliberately does **not** fall back to minting - that would
defeat revocation and split the estate into home-derived and
locally-minted principals depending on when each user first logged in.

## Decisions worth flagging

- **Provisional users carry a UUID from lookup, not from `admit`.** The
  brief specified a null UUID minted inside `admit`. Keycloak derives
  the federated storage id from the UUID and re-reads the user by that
  id later in the same login flow, so a null would break the flow
  *after* `admit` had already written. Minting early is safe: nothing is
  persisted until the KDC confirms the password. `admit` still accepts
  null and mints, so the interface matches the brief.

- **403 gets its own exception type, an actionable message and a 30s
  denial cache.** The fix lives on a different cluster from the error,
  so the message names the permission and its UUID, the denied
  principal, and the cluster that denied it. Lookup runs *before*
  password validation, so a standing misconfiguration would otherwise
  cost one remote call per login attempt from anyone who can reach the
  login page. 5xx and timeouts stay uncached, since those are transient.

- **The `WriteKrb` grant went in the service-accounts dump, not
  `service-clients.yaml`.** The brief pointed at
  `deploy/templates/auth/principals/service-clients.yaml`, but that file
  declares `KerberosKey` CRs; `sv1openid`'s permission grants live in
  `acs-service-setup/dumps/service-accounts.yaml`.

- **`JaasKerberosAuthenticator.spnegoTokenFor` already derived the
  service principal from the target URI's host**, so no fix was needed
  there. It gained a `principalName()` accessor purely for the 403
  diagnostic.

- **Timeouts share one budget.** The remote resolve runs in series after
  the local miss, so its 1.5s default and `auth.timeout.seconds` (2s)
  both come out of Keycloak's 3-second hard limit on storage lookups.
  Neither may be raised without lowering the other; comments on both say
  so.

### Automated

1. `cd acs-keycloak-spi && mvn -B test` - 123 unit tests, all passing.
2. `mvn -B verify` for the Testcontainers ITs. **These did not run in
   this environment** (Testcontainers could not detect the OrbStack
   socket, so all 4 auto-skipped). They need verifying on CI or a
   machine with a detectable Docker daemon.
3. `helm lint deploy` and `helm template` with and without
   `openid.trustedRealms` set - both render correctly (empty default
   produces `value: ""`).

### The casing fix, on a live cluster

1. Deploy and log in to Keycloak with your full UPN, e.g.
   `me1ago@FACTORYPLUS.LOCAL`, rather than the short name.
2. Before this change: "invalid username or password", with
   `error="user_not_found"` in the Keycloak log. After: the login
   succeeds.

### Derived UUIDs

Confirm the documented command agrees with the code:

```sh
python3 -c "import uuid; print(uuid.uuid5(uuid.UUID(
  'cb3714c4-c85c-482a-987b-408293aa141e'), 'me1ago@AMRC-FP.SHEF.AC.UK'))"
```

Expect `12bb7b35-16c8-572d-af5f-c1f312ec4ae8`, matching the known-answer
test in `FPAuthBackedUserStoreTest`.

### Cross-realm, recommended mode (no home-cluster grant needed)

1. Confirm the foreign realm and its KDC are in `krb5.conf` on this
   cluster and that the trust works.
2. Set, in values.yaml:
   ```yaml
   openid:
     trustedRealms:
       - realm: PARTNER.EXAMPLE.ORG
   ```
3. Upgrade. Log in as `someone@PARTNER.EXAMPLE.ORG`.
4. Expect: login succeeds. The user appears in the F+ ACL editor by
   their UPN with no permissions, under the UUID the command above
   predicts.
4b. Pre-granting: compute a UUID for a user who has NOT yet logged in,
   grant it a permission, then log in as them. Expect the permission to
   be present on first login.
5. Grant them something in F+; it appears in `fp_permissions` on the
   next login, within the SPI's 60s cache TTL.
6. Log in with a wrong password and confirm no principal is created.

### Cross-realm, shared-UUID mode

7. Add `authUrl: https://auth.partner.example.org` to that entry and
   upgrade.
8. Without the `ReadKrb` grant on the home cluster, expect the login to
   fail and the Keycloak log to name `ReadKrb`, its UUID, the denied
   `sv1openid@<THIS-REALM>`, and the remote cluster.
9. Make the grant on the home cluster, wait out the 30s denial cache,
   and log in again. Expect success, with the principal UUID matching
   the one the user has at home.

## Release notes

**Cross-realm Kerberos login for Keycloak.** Users from a trusted
foreign Kerberos realm can now sign in without an administrator creating
a Factory+ principal for them first. Set `openid.trustedRealms` in
values.yaml to opt in; it is empty by default and the feature is
disabled until configured. Requires cross-realm Kerberos trust to
already exist in `krb5.conf`.

Listing just the realm name is the recommended configuration. Each
user's Factory+ principal UUID is then derived from their username, so
every cluster arrives at the same UUID for the same person without any
coordination, and administrators can grant permissions ahead of a user's
first login. The SPI README documents the one-line command for computing
a UUID.

A realm can optionally be given an `authUrl` so users reuse the
principal UUID from their home cluster instead. That mode requires a
principal to be created there for this cluster and granted `ReadKrb` -
which cannot be narrowed to specific users and confers blanket read of
the home cluster's entire identity table - and makes logins depend on
that cluster being reachable. Choose one mode before the first login
from a realm: switching later leaves users split between the two schemes
depending on when they first signed in.

**Fixed: logins with a fully qualified username failed.** Signing in as
`user@REALM` rather than just `user` was rejected with "invalid username
or password", because Keycloak lower-cased the realm before the
Factory+ lookup. Short-name logins were unaffected.

### DIFF
--- a/acs-keycloak-spi/README.md
+++ b/acs-keycloak-spi/README.md
@@ -9,22 +9,138 @@ consumers (Grafana, acs-i3x, future shims).
 See:
 - Pitch: `docs/plans/2026-05-07-keycloak-fp-user-storage-spi.md`
 - Implementation plan: `docs/plans/2026-05-07-keycloak-fp-spi-plan.md`
+- Cross-realm design: `docs/plans/2026-07-27-keycloak-spi-cross-realm-design.md`
 
 ## Status
 
-**Phase 1 of 12 — build skeleton + hello-world SPI.**
+In production since ACS v6. The plugin:
 
-The plugin currently:
-- Loads in Keycloak as the `factoryplus` user storage provider
-- Implements `UserLookupProvider` against an injectable `FactoryPlusUserStore`
-- Ships with a `NullFactoryPlusUserStore` default that returns no users
+- Loads in Keycloak as the `factoryplus` user storage provider, and is
+  provisioned as the realm's federation by `acs-service-setup`
+- Resolves users from the Factory+ auth service over its v2 identity API,
+  authenticating with SPNEGO as `sv1openid`
+- Validates passwords against the KDC via `Krb5LoginModule`
+- Stamps `fp_principal_uuid` and `fp_permissions` into issued tokens via
+  its two protocol mappers
+- Caches hits and misses for a configurable TTL
+- Accepts logins from configured foreign Kerberos realms (off by default;
+  see Cross-realm login below)
 
-It does NOT yet:
-- Talk to a real Factory+ auth service (Phase 2)
-- Implement group lookups, credential validation, or claim mappers
-- Replace Kerberos federation in the Helm deployment
+It does not:
+- Support email lookup (Factory+ has no email field)
+- Support free-text user search, arbitrary user attributes, or group
+  membership queries
+- Write to Factory+, other than mirroring a cross-realm user's Kerberos
+  identity after their password has validated
 
-Refer to the plan doc for the full phase map.
+## Configuration
+
+Set on the federation component. `acs-service-setup/lib/openid.js`
+populates all of these from the Helm chart, so on a normal ACS
+deployment you configure them through `values.yaml` rather than by hand.
+
+| Key | Default | Meaning |
+|---|---|---|
+| `auth.url` | (blank) | Base URL of the Factory+ auth service. Blank disables F+ lookups entirely: the federation loads but returns no users. |
+| `auth.timeout.seconds` | `2` | Per-request timeout for F+ auth calls. **Must stay under 3s**, see below. |
+| `cache.ttl.seconds` | `60` | How long to cache lookups, including misses. `0` disables caching. |
+| `auth.principal` | (blank) | Principal the SPI authenticates as, e.g. `sv1openid@FACTORYPLUS.LOCAL`. Blank calls F+ unauthenticated, which is only useful against a stand-in server. |
+| `auth.keytab.path` | (blank) | Keytab holding credentials for that principal. Must be readable by Keycloak's process. |
+| `default.realm` | (blank) | Realm appended to usernames typed without an `@realm` suffix, so local users can log in with a short name. |
+| `trusted.realms` | (blank) | Comma-separated Kerberos realms accepted for cross-realm login. Blank means none. |
+| `trusted.realm.auth.urls` | (blank) | Comma-separated `REALM=url` entries naming each trusted realm's own F+ auth service. |
+| `trusted.realm.timeout.seconds` | `1.5` | Per-request timeout for calls to a trusted realm's auth service. |
+
+### The 3-second budget
+
+Keycloak hard-kills user storage lookups at 3 seconds, and overshooting
+surfaces as an opaque `InterruptedException` rather than a timeout
+naming the slow service. The cross-realm resolve runs **in series**
+after the local lookup misses, so `auth.timeout.seconds` and
+`trusted.realm.timeout.seconds` share that one budget. Do not raise
+either without lowering the other.
+
+### Username casing
+
+Keycloak lower-cases the username before it reaches the provider. The
+SPI tries the resulting UPN verbatim first, then retries with the realm
+portion (everything after the last `@`) upper-cased. That covers both
+the usual uppercase realms and deployments whose realm genuinely is
+lower case.
+
+Only the realm is touched. An F+ identity stored with capitals in the
+**user** portion (`Me1ago@REALM`) cannot be matched, because the
+original casing is gone by the time we see it. Store user portions in
+lower case.
+
+## Cross-realm login
+
+With `trusted.realms` set, a user from a listed realm can sign in with
+no pre-existing Factory+ principal on this cluster. Once their home KDC
+accepts the password, the SPI writes their Kerberos identity here and
+they appear in the ACL editor by their UPN, with no permissions until an
+admin grants some. Nothing is written before the KDC has confirmed the
+password.
+
+This is off by default and requires cross-realm Kerberos trust to
+already exist: the realm and its KDC must be in `krb5.conf` on this
+cluster. Listing a realm here does not create that trust.
+
+There are two supported modes, and the choice matters.
+
+### With an `authUrl`: shared principal UUID, blanket read on the home cluster
+
+Give the realm an entry in `trusted.realm.auth.urls` and the SPI
+resolves the user's principal UUID from their home cluster, so they keep
+one identity across both.
+
+This requires a manual grant **on the home cluster**: the consuming
+cluster's `sv1openid@<CONSUMING-REALM>` must hold `ReadKrb`
+(`e8c9c0f7-0d54-4db2-b8d6-cd80c45f6a5c`).
+
+Know what you are granting. `ReadKrb` cannot be scoped to individual
+principals - it can only be granted on Wildcard, and acs-auth treats it
+as a blanket "read any identity" capability (see the comment at
+`acs-auth/lib/dataflow.js:273`). The grant lets the consuming cluster's
+Keycloak **read and enumerate every identity record in the home
+cluster's auth service**, as a standing capability, regardless of who
+actually logs in or whether anyone ever does.
+
+That grant is also the revocation point for the whole trust
+relationship: remove it and the consuming cluster can no longer resolve
+home principals.
+
+Without the grant the home cluster returns 403 and the login **fails**.
+It does not quietly fall back to minting a local UUID - that would
+defeat revocation, and would leave you with some users on home-derived
+UUIDs and some on local ones depending on when they first logged in.
+The logged message names the missing permission, the principal that was
+denied and the cluster that denied it. A standing denial is cached for
+30 seconds so it costs one outbound call rather than one per login
+attempt; 5xx and timeouts stay uncached, since those are transient and
+recovery should be immediate.
+
+### Without an `authUrl`: no grant, no outbound call, per-cluster UUID
+
+Listing a realm in `trusted.realms` with no entry in
+`trusted.realm.auth.urls` is a fully supported configuration, not a
+degraded fallback. On this path the consuming cluster makes no outbound
+call to the home cluster at all and needs no permission there
+whatsoever; it mints a fresh local principal UUID on first successful
+login.
+
+The cost is that the UUID is not shared: the user is recognisably the
+same person by UPN, but each cluster's auth service knows them by a
+different UUID, so grants do not correlate across clusters.
+
+If you are unwilling to grant blanket identity read across a cluster
+boundary, use this mode.
+
+### On this cluster
+
+`sv1openid` needs `WriteKrb` (`327c4cc8-9c46-4e1e-bb6b-257ace37b0f6`) on
+Wildcard to mirror the identity. The Helm chart grants this
+automatically via `acs-service-setup/dumps/service-accounts.yaml`.
 
 ## Build
 
@@ -44,9 +160,8 @@ mvn -B test       # unit tests (no Docker required)
 mvn -B verify     # unit + integration tests (Testcontainers, requires Docker)
 ```
 
-The current Phase 1 test count is 17 unit tests (5 classes) + 2
-integration tests in `FactoryPlusFederationIT` (real Keycloak via
-Testcontainers).
+Currently 115 unit tests across 10 classes, plus 4 integration tests in
+`FactoryPlusFederationIT` (real Keycloak via Testcontainers).
 
 ### Running integration tests locally
 
--- a/acs-keycloak-spi/README.md
+++ b/acs-keycloak-spi/README.md
@@ -86,17 +86,62 @@ This is off by default and requires cross-realm Kerberos trust to
 already exist: the realm and its KDC must be in `krb5.conf` on this
 cluster. Listing a realm here does not create that trust.
 
-There are two supported modes, and the choice matters.
+There are two modes. Pick deliberately, and see the one-way-door warning
+at the end.
 
-### With an `authUrl`: shared principal UUID, blanket read on the home cluster
+### Recommended: no `authUrl`, derived UUIDs
+
+List the realm and nothing else. The principal UUID is derived from the
+user's UPN as a UUIDv5, which means:
+
+- No principal to create and no permission to obtain on the home
+  cluster. No coordination with whoever runs it.
+- No outbound call at login, so no dependency on the home cluster being
+  up, reachable, or still existing.
+- Every ACS cluster derives the *same* UUID for the same person, with no
+  coordination, so identity stays consistent across the estate anyway.
+- **Admins can pre-grant**: compute the UUID and attach permissions
+  before the user has ever logged in, so their first login works
+  properly instead of landing with nothing.
+
+What you give up is reusing the UUID the user has *on their home
+cluster*, so grants made there do not follow them here.
+
+#### Computing a UUID to pre-grant
+
+```sh
+python3 -c "import uuid; print(uuid.uuid5(uuid.UUID(
+  'cb3714c4-c85c-482a-987b-408293aa141e'), 'me1ago@AMRC-FP.SHEF.AC.UK'))"
+```
+
+```
+12bb7b35-16c8-572d-af5f-c1f312ec4ae8
+```
+
+The name you hash is the **canonicalised UPN**: realm portion in
+CAPITALS, user portion exactly as the person types it into the login
+form (which Keycloak lower-cases, so in practice lower case). Nothing
+local feeds into it - not the default realm, not the cluster name - and
+that is what makes every cluster agree.
+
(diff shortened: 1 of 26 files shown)


AlexGodbehere:
# Review: cross-realm Kerberos login for the Keycloak SPI, plus a realm-casing bug fix

**Branch:** `agent-a4609cbe9b133e861`
**Worktree:** `.claude/worktrees/agent-a4609cbe9b133e861`
**PR:** #698

Two things, deliberately in separate commits because the first stands
alone as a bug fix.

**The casing bug.** Keycloak lower-cases the username before handing it
to a user storage provider. Factory+ Auth compares identity names with
exact string equality. So a user signing in as
`me1ago@AMRC-FP.SHEF.AC.UK` had `me1ago@amrc-fp.shef.ac.uk` looked up,
got a 410, and saw "invalid username or password" with
`error="user_not_found"` in the Keycloak log. This was not cross-realm
specific: any local user typing their full UPN instead of their short
name hit it. Short names worked only because `default.realm` is appended
*after* Keycloak has finished folding the case.

The lookup now builds a candidate list and tries them in order:

| Typed | Candidate 1 | Candidate 2 |
|---|---|---|
| `me1ago` (short name, default.realm=FACTORYPLUS.LOCAL) | `me1ago@FACTORYPLUS.LOCAL` | (none, already upper) |
| `me1ago@AMRC-FP.SHEF.AC.UK` (arrives lower-cased) | `me1ago@amrc-fp.shef.ac.uk` | `me1ago@AMRC-FP.SHEF.AC.UK` |
| `alice@mixed.Realm` on a genuinely-lowercase-realm cluster | `alice@mixed.Realm` (hits) | not attempted |

Retry rather than unconditional upper-casing is the point: a deployment
whose realm really is lower case keeps working and pays no second
request. Only the realm is touched, so an identity stored as
`Me1ago@REALM` is still unreachable - Keycloak folded that case before
we saw it and it cannot be recovered.

**Cross-realm login.** Kerberos cross-realm trust is a KDC-level fact.
Adding a realm to `krb5.conf` lets the AS-REQ route to the foreign KDC
but provisions nothing in Factory+, so a foreign user has no principal
here and the login dies before the password is tested. Previously an
admin had to hand-create a principal first.

Now, when every casing candidate misses locally *and* the realm is in
`trusted.realms`, the lookup returns a **provisional** user. If the
password then validates against the home KDC, the provider writes their
Kerberos identity here (`PUT /v2/principal/{uuid}/kerberos`) and the
session is issued. They appear in the ACL editor by their UPN with zero
permissions until granted some. If the password fails, nothing is
written at all.

Default config trusts nothing, so all of this is inert until configured.

**Derived UUIDs.** On the recommended path (realm listed, no `authUrl`)
the principal UUID is not random - it is a UUIDv5 over a fixed namespace
and the canonicalised UPN. So every cluster independently derives the
same UUID for the same person, and an admin can compute it and grant
permissions *before* the user's first login. Worked example:

```sh
python3 -c "import uuid; print(uuid.uuid5(uuid.UUID(
  'cb3714c4-c85c-482a-987b-408293aa141e'), 'me1ago@AMRC-FP.SHEF.AC.UK'))"
# 12bb7b35-16c8-572d-af5f-c1f312ec4ae8
```

That pair is a known-answer test, so the code cannot drift from the
documented command.

## Walkthrough

### Realm casing
- Fully qualified logins work now, for local users as well as foreign ones.
- A hit on the verbatim candidate costs no extra request; the retry only
  fires on a miss. Tested explicitly, because both timeouts come out of
  Keycloak's 3s budget.
- `candidateUpns` is package-visible and unit-tested directly, so the
  "only the realm, never the user portion" rule is pinned.
- Fixed the comment at `KerberosPasswordValidator:37` which claimed
  "same realm-uppercase canonicalisation as the user lookup". That was
  false before this change: the validator upper-cased, the lookup didn't.

### Trusted realms, two modes
- `trusted.realms` is a comma-separated list, matched case-insensitively
  (the incoming realm is lower-cased, the configured one may be typed
  either way). Empty by default.
- **Recommended: no `authUrl`.** UUID derived from the UPN, no outbound
  call at all, no grant to obtain, no dependency on the home cluster.
  Now the documented default.
- **With an `authUrl`**, the principal UUID is resolved from that realm's
  own F+ Auth, so the user carries that cluster's UUID here. Needs a
  principal created for our `sv1openid` on the home cluster plus a
  `ReadKrb` grant there.
- **Switching later is close to a one-way door**: users who already
  logged in keep their derived UUID while first-time users get
  home-derived ones, splitting the estate by when each person first
  signed in. Documented in values.yaml, the README and the design doc.
- Remote stores are built once at construction, keyed by realm, because
  each one does a JAAS login and per-request construction would blow the
  storage-lookup budget.

### UUID derivation
- UUIDv5, RFC 4122 name-based SHA-1. Hand-rolled: the JDK has no v5
  generator and `UUID.nameUUIDFromBytes` is MD5/v3, not a substitute.
  Verified against python's `uuid.uuid5`.
- Namespace `cb3714c4-c85c-482a-987b-408293aa141e` is fixed forever;
  changing it orphans every principal ever derived from it. Pinned by a
  test, as is a known-answer vector.
- Name input is the canonicalised UPN and **nothing local feeds into
  it**. If it varied with local config, clusters would disagree and the
  whole property would be lost. There is a comment saying so.
- **Mint path only.** A UUID resolved from a home cluster is adopted
  verbatim, with a test asserting the remote path does not derive.
- `trusted.realm.auth.urls` uses a flat `REALM=url` encoding because
  Keycloak component config is a fixed set of keys declared in
  `getConfigProperties()`; dynamic per-realm keys aren't expressible.

### The `ReadKrb` grant, and why the mint-fresh mode matters
- `ReadKrb` **cannot be scoped to individual principals**. It is
  Wildcard-only and acs-auth treats it as blanket "read any identity"
  (`acs-auth/lib/dataflow.js:273`).
- So granting it is not "let them look up the users who log in there".
  It is "let that cluster's Keycloak read and enumerate every identity
  record in this cluster's auth service", standing, regardless of who
  logs in or whether anyone ever does.
- It is also the revocation point for the whole trust relationship.
- The no-`authUrl` mode needs no grant and makes no outbound call. Both
  the design doc and the README present it as the option for anyone
  unwilling to grant blanket identity read across a cluster boundary.
  Cost: per-cluster UUIDs, so grants don't correlate.

### Write ordering, which is the security property
- `admit` is called from exactly one place, after
  `passwordValidator.validate` has returned true. Tested explicitly,
  including that it is *never* called on a failed password.
- If `admit` throws, the login is rejected. Issuing the session anyway
  would hand out a token whose `fp_principal_uuid` nothing on the
  cluster recognises - every F+ service would reject it and the user
  would see a working login with no access and no explanation.
- The provider detects provisional users by `instanceof` with an
  `fp_provisional` attribute fallback, because by credential-validation
  time we may be behind one of Keycloak's own `UserModel` wrappers.
  That's the same reason `fp_principal_uuid` is exposed as an attribute.
  No mapper reads `fp_provisional`, so it never reaches a token.

### Error handling, and the deliberate caching asymmetry
- **410/404**: user genuinely absent at home. Empty, negatively cached,
  login rejected.
- **5xx / timeout**: throws. Keycloak fails the login rather than
  reporting `user_not_found` - "the home cluster is down" and "no such
  user" must not collapse into one answer. **Not cached**, so recovery
  is immediate.
- **403**: the home cluster hasn't granted us `ReadKrb`. Fatal, and it
  deliberately does *not* fall back to minting - that would defeat
  revocation and split the estate into home-derived and locally-minted
  principals depending on when each user first logged in. **Cached 30s**,
  keyed by realm, because lookup runs *before* password validation, so a
  standing misconfiguration would otherwise cost one remote call per
  login attempt from anyone who can reach the login page. The cache
  suppresses the call without softening the failure.
- The 403 message names the permission and its UUID, the denied
  principal, and the cluster that denied it, since the fix lives on a
  different cluster from the error.

### Deployment
- `openid.trustedRealms` in values.yaml, default `[]`, rendered flat by
  the chart into two env vars on the service-setup job.
- `sv1openid` gains `Auth.Perm.ManageKerberos` (WriteKrb) on Wildcard,
  matching `op1krbkeys`. With `trustedRealms` empty the endpoint is
  never called.

## How to test

1. `cd acs-keycloak-spi && mvn -B test` - expect 123 passing.
2. `mvn -B verify` on a machine with a Testcontainers-detectable Docker
   daemon. **Not verified here** - all 4 ITs auto-skipped because
   Testcontainers couldn't detect the OrbStack socket.
3. `helm lint deploy`, and `helm template` with and without
   `openid.trustedRealms` set.
4. On a live cluster, log in with your full UPN (`user@REALM`) rather
   than the short name. Before: rejected. After: succeeds.
5. Set `openid.trustedRealms: [{realm: PARTNER.EXAMPLE.ORG}]` (no
   authUrl), upgrade, and log in as a partner-realm user. Expect
   success, and the user visible in the ACL editor by UPN with no
   permissions, under the UUID the python one-liner predicts.
5b. Pre-grant: compute the UUID for a user who has never logged in,
   attach a permission to it, then log in as them. Expect the permission
   present on first login - this is the main thing the derivation buys.
6. Log in with a deliberately wrong password and confirm no principal
   was created.
7. Add an `authUrl` without making the `ReadKrb` grant on the home
   cluster. Expect the login to fail and the log to name `ReadKrb`, its
   UUID, `sv1openid@<THIS-REALM>` and the remote cluster.
8. Make the grant, wait out the 30s denial cache, log in again. Expect
   success with the home cluster's UUID.

## Decisions / open questions

- **Provisional users carry a UUID from lookup, not from `admit`.** The
  brief specified null-until-admit. Keycloak derives the federated
  storage id from the UUID and re-reads the user by that id later in the
  same login flow, so a null would break the flow *after* `admit` had
  written. Minting early is safe - nothing is persisted until the KDC
  confirms the password. `admit` still accepts null and mints, so the
  interface matches the brief; that branch is defensive only.
- **The `WriteKrb` grant went in the service-accounts dump.** The brief
  pointed at `deploy/templates/auth/principals/service-clients.yaml`,
  but that file declares `KerberosKey` CRs, not grants. `sv1openid`'s
  grants live in `acs-service-setup/dumps/service-accounts.yaml`.
- **No fix needed to `spnegoTokenFor`.** The brief flagged it as
  possibly hardcoding the local host; it already derives the service
  principal from the target URI's host. It only gained a
  `principalName()` accessor for the 403 diagnostic.
- **Integration tests unverified locally.** Testcontainers wouldn't
  detect the OrbStack socket even with `DOCKER_HOST` set and the daemon
  reachable via the CLI - pre-existing friction, documented in the SPI
  README. CI should confirm.
- **Deferred:** the mirrored principal gets no ConfigDB `Principal`
  object or `Info` name, so it shows in the ACL editor by UPN rather
  than a friendly name. Adding one means giving the SPI a ConfigDB
  client it doesn't have. Recorded as follow-up in the design doc.
- **Accepted risk:** auth-DB growth from cross-realm users who log in
  once and are never granted anything. No reaper, by decision.
- **Settled:** the namespace UUID
  `cb3714c4-c85c-482a-987b-408293aa141e` is signed off. It is a
  permanent commitment - it cannot change later without orphaning every
  principal derived from it - so it is pinned by a test and flagged in
  the code and design doc. Not an open question.

## Files of note

**The casing fix (commit 1)**
- `acs-keycloak-spi/.../FPAuthBackedUserStore.java` - `candidateUpns`
- `acs-keycloak-spi/.../KerberosPasswordValidator.java` - corrected comment

**UUID derivation (commit 5)**
- `acs-keycloak-spi/.../FPAuthBackedUserStore.java` - `uuidV5`,
  `mintedUuid`, `ACS_PRINCIPAL_NAMESPACE`

**Cross-realm core (commit 2)**
- `acs-keycloak-spi/.../FPAuthBackedUserStore.java` - `resolveCrossRealm`,
  `admit`, the denial cache
- `acs-keycloak-spi/.../FactoryPlusUserStorageProvider.java` - admit-after-validate
- `acs-keycloak-spi/.../CachingFactoryPlusUserStore.java` - invalidation on admit
- `acs-keycloak-spi/.../FactoryPlusAccessDeniedException.java` - new, 403 only
- `acs-keycloak-spi/.../FactoryPlusUserStorageProviderFactory.java` - config keys

**Deployment (commit 3)**
- `deploy/values.yaml` - `openid.trustedRealms`
- `deploy/templates/service-setup.yaml` - flat env rendering
- `acs-service-setup/lib/openid.js` - component config mapping
- `acs-service-setup/dumps/service-accounts.yaml` - WriteKrb grant

**Docs (commit 4)**
- `docs/plans/2026-07-27-keycloak-spi-cross-realm-design.md`
- `acs-keycloak-spi/README.md`

PR #702, opened 2026-07-30
Schema editor: author metric schemas in the admin UI

AlexGodbehere:
Brings back an SME-facing schema editor, redesigned rather than restored. Shaped as a Shape Up pitch first; the pitch is in `docs/plans/2026-07-30-schema-editor.md`.

## The problem

An engineer who knows what a machine measures cannot express it. The only authoring path is hand-written YAML PR'd into the AMRC `acs-schemas` repo, which is shared upstream and the wrong place for a site's own machine definitions. Days of round-trip for a schema only one site will ever use.

There was an editor in `acs-manager`, deleted in `d5ab3d90`. It downloaded a JSON file and stopped, addressed schemas by GitHub URL, and rendered anything it did not recognise as the literal string `BAD FORMAT`.

## What this does

The ConfigDB is the source of truth for a deployment's schemas. The git library is a starting point, not an authoring channel. Nothing here writes to git.

Two rules, the first forced by existing code:

- **Library schema**: never edited in place. `acs-git`'s schema hook re-writes any schema whose stored `source` matches the repo it pulled from, so an in-place edit is reverted on the next pull. Editing forks it.
- **Local schema**: additive edits update in place, breaking edits fork to a new UUID and bump the version.

Which of those happens is decided by a change classifier, not chosen by the author.

| Change | Class |
|---|---|
| Add a metric, component, component list or group | additive |
| Change documentation, unit, range, historian flag | additive |
| Add an allowed Sparkplug type | additive |
| Remove or rename a node | breaking |
| Remove an allowed Sparkplug type | breaking |
| Point a component at a different schema | breaking |
| Add a `required` entry | breaking |
| Change anything inside a node the composer does not model | breaking |
| Reorder nodes | not a change |

## The part worth reviewing carefully

**The document model is lossless.** Every leaf keeps the JSON it was parsed from; serialising patches only recognised fields. Anything the composer does not model becomes an opaque node and round-trips byte for byte. This is what makes the raw JSON view safe and why there is no `BAD FORMAT` state.

The test asserts parse-then-serialise is byte-identical across **all 139 schemas in the current `acs-schemas` library**. It caught three real bugs during the build: injecting `$schema` into schemas that omit it, dropping metric fields whose value is a legitimate `""` or `null`, and adding `type: object` to component lists that do not declare it. Each would have silently rewritten files nobody had touched.

**The classifier does not diff arbitrary JSON Schema.** That is an open problem and would have eaten the appetite. It diffs the composer's closed vocabulary and treats any change touching an opaque node as breaking, without analysis.

## Blast radius needed no new API

Both figures come from data already in the browser or already served:

- **Configured**: walks the device store's origin maps for the schema at any depth, including nested components. The admin UI already loads every device's `DeviceInformation`.
- **Publishing now**: Directory's existing `GET /v1/schema/{uuid}/devices`. Reported as *unknown* rather than *zero* when the Directory cannot be reached.

## Known issue, pre-existing, not fixed here

**`acs-admin` cannot be built from a clean local tree, on this branch or on `main`.** `lib/js-service-client` imports `long` without declaring it, and its other declared dependencies do not reach the consumer through the `file:` link either. Verified by building `main` at `2b5358e9` against the same modules: it fails identically.

The Docker build works because it bind-mounts `lib/` so bun sees a real workspace. To build locally:

```sh
cd acs-admin
npm install --no-save long content-type isomorphic-ws mqtt optional-js semver sparkplug-payload
bun run build
```

I started to fix it and backed out: adding one of the six declarations is worse than adding none, and a proper fix touches a shared library that other services consume. Worth its own PR.

## How to test

1. `cd acs-admin && bun install && npm test` — 359 tests, including the round-trip over the real library.
2. Deploy and open **Schemas** in the admin UI. The AMRC library appears, marked as such.
3. Select a library schema, for example CNC. **Make a local copy**. It requires a new name and starts at v1.
4. In the composer, add a metric. Set its Sparkplug type and unit. **Save**, then **Publish**. It publishes as v1.
5. Edit it again and add a second metric. **Publish**. The gate reports the change as safe and updates in place; the version does not move.
6. Now rename an existing metric. **Publish**. The gate reports it as breaking, shows how many devices are affected, and the only action is *Publish as v2*.
7. Back in the list, v1 is marked superseded. Any device on v1 shows a "v2 available" link on its device page.
8. Open a hand-written schema with an unusual construct, check it renders read-only rather than erroring, then **Raw** and confirm the JSON is unchanged.

## Release notes

The admin UI can now author metric schemas. Schemas live in the ConfigDB and the AMRC schema library becomes a starting point rather than the only source. Editing a library schema makes a local copy; editing your own updates it in place when the change is safe and creates a new version when it is not, so a change can never break a device that is already running. Before publishing you are shown what changed and how many devices it affects.

## Still to come

Visual design is deliberately plain. A design brief for the UI pass is in `docs/plans/2026-07-30-schema-editor-design-brief.md`; that work is expected to land on top of this.

Out of scope by decision, recorded in the pitch: deriving schemas from live devices, assisted migration of devices between versions, writing back to the AMRC library, and a schema-author role below Administrator.


AlexGodbehere:
# Review: SME-facing schema editor, with lossless editing and a change classifier

**Branch:** `schema-editor`
**Worktree:** `.claude/worktrees/schema-editor`
**PR:** #702

An engineer can now author a Factory+ metric schema from the admin UI. Previously the only path was hand-written YAML PR'd into the shared AMRC `acs-schemas` repo.

The model: the ConfigDB owns a deployment's schemas, the git library is a starting point. Two rules follow, and the first is forced by code that already exists.

`acs-git/lib/hooks/schemas.js` re-writes any schema whose stored `source` matches the repo it pulled from. So an in-place edit of a library schema is silently reverted on the next pull. Editing a library schema therefore forks it. A local schema follows the classifier: additive edits update in place, breaking edits fork and bump the version.

Worked example. A site forks the library's `CNC v1`, names it `CNC Sheffield`, and it publishes at v1 with `derivedFrom` pointing at the library CNC.

- They add `Coolant_Temp`. Classified additive. Publishing updates `CNC Sheffield v1` in place. Four devices already using it are untouched.
- They rename `Feed` to `Feedrate`. Classified breaking. Publishing creates a new UUID at v2 with `replaces` pointing at v1. The four devices stay on v1, which still exists unchanged. v1 gains a superseded badge, and each of those devices shows "v2 available".

The fork is named rather than versioned deliberately. Had it become `CNC v2`, the day AMRC ships a real `CNC v2` there would be two different schemas with the same name and version.

## Walkthrough

### Lossless document model

- `document.js` parses into four node kinds: metric, component, component list, group. Anything else becomes an opaque node held verbatim.
- Every leaf keeps the JSON it was parsed from. Serialising patches only recognised fields, so untouched parts are emitted byte for byte.
- The test asserts byte-identical round-trip across **all 139 schemas in the live `acs-schemas` library**. It caught three real bugs: `$schema` injected into schemas that omit it, metric fields dropped when their value is a legitimate `""` or `null`, and `type: object` added to component lists that do not declare it.
- The consequence is that the raw JSON view is safe to ship editable, and there is no `BAD FORMAT` state. The old editor had one.

### Change classifier

- Diffs the composer's closed vocabulary, not arbitrary JSON Schema. General semantic diffing of JSON Schema is an open problem and would have consumed the appetite.
- Any change touching an opaque node is classified breaking with no analysis attempted.
- Narrowing a metric's allowed Sparkplug types is breaking; widening is not. Reordering is not a change, since property order carries no meaning.
- Renames are detected structurally (a removal and an addition with identical bodies) so the change list reads "Feed renamed to Feedrate" rather than as an unrelated delete and create. Both classify breaking either way.

### Publish gate

- Shows the classified changes split into breaking and safe, then the reach, then acts. The outcome is a consequence of the classification, not a choice offered to the author.
- Blocked while the schema references a component that is still a draft. The origin map editor's ref resolver does `found.schema` with no guard, so an unresolvable reference throws and takes that editor down.

### Blast radius, no new API

- **Configured** walks the device store's origin maps for the schema at any depth. The admin UI already holds every device's `DeviceInformation`, so nested component use is found, which a top-level config search would miss.
- **Publishing now** uses Directory's existing `GET /v1/schema/{uuid}/devices`.
- The two are kept apart and reported separately. When the Directory cannot be reached the live figure is *unknown*, never *zero*.

### Drafts

- Held in their own class, deliberately not `Class.Schema`. The schema store, origin map editor, acs-i3x and the edge agent all enumerate by that class membership, so a half-built schema is invisible to all of them.

## How to test

1. `cd acs-admin && bun install && npm test` — 359 tests pass, including the library round-trip.
2. Deploy, open **Schemas**. The AMRC library is listed and marked.
3. Select a library schema, **Make a local copy**. Confirm it demands a new name and starts at v1.
4. Add a metric, set type and unit, **Save**, **Publish**. Publishes as v1.
5. Add a second metric, **Publish**. Reported safe, updates in place, version unchanged.
6. Rename an existing metric, **Publish**. Reported breaking, shows affected device count, only offers *Publish as v2*.
7. Confirm v1 is now badged superseded, and a device using v1 shows "v2 available".
8. Open a hand-written schema with an unusual construct. Confirm it renders read-only rather than erroring, and **Raw** shows the JSON unchanged.

## Decisions / open questions

- **`acs-admin` does not build from a clean local tree, on this branch or on `main`.** `lib/js-service-client` imports `long` without declaring it, and its declared deps do not reach the consumer through the `file:` link either. Verified by building `main` at `2b5358e9` against the same modules; it fails identically. Docker works because it bind-mounts `lib/`. I started fixing it and backed out: one of six declarations is worse than none, and a real fix touches a shared library. Wants its own PR.
- **Reordering is presentation, not a breaking change.** You agreed during shaping. JSON Schema property order carries no meaning and consumers key by name.
- **"Component" collides with "component" meaning an ACS service** in the docs. Kept because it reads better for the audience, with a one-line definition where it first appears.
- **Visual design is deliberately plain.** The brief for the design pass is in `docs/plans/2026-07-30-schema-editor-design-brief.md`.
- `acs-admin` had no test framework. Added vitest, following `acs-data-access`.

## Files of note

**The load-bearing layer**
- `acs-admin/src/lib/schema/document.js` — lossless parse and serialise
- `acs-admin/src/lib/schema/classify.js` — the additive/breaking table
- `acs-admin/src/lib/schema/lineage.js` — provenance and version rules
- `acs-admin/src/lib/schema/usage.js` — blast radius

**Proof it works**
- `acs-admin/test/schema-document.test.js` — round-trip over the real library
- `acs-admin/test/schema-classify.test.js` — the classification table

**UI**
- `acs-admin/src/pages/Schemas/SchemaEditor.vue` — the composer
- `acs-admin/src/components/Schemas/PublishDialog.vue` — the gate

**Wiring**
- `acs-service-setup/dumps/schema-drafts.yaml` — draft class and app
- `docs/services/schema-editor.md` — the rules, written down

AlexGodbehere:
## Screenshots

Rendered from the real components against fixture data via the preview harness in `acs-admin/preview` (`bun run dev`, then `/preview/index.html`). Every figure below is genuine: the change lists come from running the classifier, and the device counts from walking fixture origin maps.

**Visual design is deliberately plain.** This is here to be compared against the design pass, not defended.

### Schema list

Provenance, usage and lineage across library, local and draft schemas. Note `CNC Sheffield` appears twice, v1 badged superseded by v2.

Schema list

Selecting a row shows lineage in both directions, forked from and superseded by.

Schema list with selection

### Composer

A draft mid-authoring: metrics, a group, a component, a component list, and the two reserved properties.

Composer

The metric panel. **The 28-value Sparkplug type selector is the weakest thing here** and is called out in the design brief.

Metric panel

A node the composer does not model. Held verbatim, shown read-only, and it must read as deliberate rather than as an error, because the user has done nothing wrong.

Opaque node

A component list, with the naming pattern that constrains instance names.

Component list

A library schema opens read-only. The only action is to make a local copy.

Library schema read-only

### Publish gate

The hardest screen. A breaking change forces a new version; it is not offered as a choice. The two device figures mean different things and are kept apart.

Publish, breaking

The same screen when every change is additive. It updates in place and the version does not move.

Publish, additive

Blocked because a component is still a draft. The remedy names the schema to publish first.

Publish, blocked

### Component picker and raw view

"A component is another schema used inside this one" is the one line that teaches the term, placed where it first appears.

Component picker

The escape hatch. Safe to ship editable because the document model is lossless.

Raw schema

AlexGodbehere:
## Redesigned

The design from the Claude Design project is now implemented. Screenshots below replace the earlier plain ones.

Implemented in the app's own idiom rather than by vendoring CSS: the design system in that project is a transcription of acs-admin's existing shadcn-vue components and Tailwind slate palette, so every `acs-*` class maps back to a component already in the tree, and every token is a Tailwind value the app already uses.

### The two ideas the design argues for

**Publish is not a confirmation.** The author has no decision to make, since the classifier decides whether the schema updates in place or forks. So it is a page now rather than a dialog, leading with the outcome in one sentence and following with the evidence.

Publish, breaking

The quiet case stays quiet. Same page, no warning furniture.

Publish, additive

**Reserved properties leave the tree.** `Schema_UUID` and `Instance_UUID` become a platform strip above it, so everything in the tree is something the user can actually edit.

Editor

### Type selector

The 28-value wall becomes a searchable popover with grouped types, range hints, and selected types pinned at the bottom.

Type selector

The design showed simplified names (Float, Int32). The real Sparkplug list is used instead, and a test asserts the grouping covers all 28 exactly once, so a type cannot silently become unreachable.

### Conventions

Colour means something now: green additive, amber breaking, red reserved for failure. A breaking change is serious, not an error. Origin reads as ink weight before it reads as a badge, so library rows sit muted and locked while local rows are full weight. Published carries no badge, because it is the default state.

Schema list

Schema list, all

### Other states

A node the composer does not model, reading as preserved rather than broken:

Not modelled

Publishing blocked on an unpublished component, and the live count when the Directory is unreachable:

Publish blocked

Live count unknown

A library schema opens read-only, metric panel, component panel, picker and raw view:

Library read-only

Metric panel

Component panel

Component picker

Raw schema

### One departure from the design

A component list showed an instance count ("Spindles 3"). A schema constrains what instances may be *named* but does not enumerate them; only a device's origin map knows how many exist. It reads "many" rather than a fabricated number.

### Unchanged

The document model, classifier, lineage and usage layers are untouched. That was the point of separating them: this was a UI change and the 372 tests, including the byte-identical round-trip over all 139 library schemas, pass without modification.

PR #714, opened 2026-08-11
Custom driver UIs in the Manager, and the Pathfindr driver

AlexGodbehere:
Delivers two shaped pitches in one PR: a mechanism letting any driver ship its own configuration interface, and the Pathfindr driver that is its first consumer. Pitch A lands before Pitch B in the commit history.

Pitches are included in the diff at `docs/plans/2026-08-11-custom-driver-uis.md` and `docs/plans/2026-08-11-pathfindr-driver.md`.

## Why

The Pathfindr API has no address strings in it. It has a choice of data type and an asset serial, two structured inputs. The only thing the Manager offers is a free-text box labelled "Device Address", so a driver author has to invent a grammar like `assets/SN12345`, document it in a README, and hope the operator reads it.

Every driver does this. ADS invented `symbol_name,cycle_time`, RTDE invented `json/jointData`, the test driver invented `sin:3000:50:bd`. None of these exist because they are good ways to identify anything. They exist because a text box was the only affordance available. The address string is a workaround for the absence of a UI.

## Pitch A: custom driver UIs

A driver definition can carry a `ui` key holding a self-contained HTML document. The Manager renders it in place of the per-metric fields.

**Scope is deliberately narrow.** It replaces only the per-metric region: address, path, Sparkplug type, engineering limits. The tag tree, schema selection, static values and the CSV path are untouched. The origin map editor is around 3,100 lines across seven components, and a feature that asked driver authors to reimplement it would be used exactly once, by us.

**Delivery adds no infrastructure.** A third-party driver already has to get its definition into ConfigDB, and the deploy chart already supports that via the `EXTRA_DUMPS` ConfigMap mounted at `/dumps`. Shipping a UI is one more key in a file the author was already writing. No file uploads, no image extraction.

**Containment is two independent mechanisms.** The iframe carries `sandbox="allow-scripts"` and deliberately not `allow-same-origin`, so the document gets an opaque origin and cannot reach the console, its storage or the operator's session. Separately, the frame is served with `default-src 'none'`, so it cannot fetch, XHR, open a WebSocket or load a subresource. "A driver UI cannot call the network" is enforced by the browser rather than documented as a convention.

Everything the document sends back is validated against an allowlist before it reaches the metric model, and the Manager remains the only thing that writes to ConfigDB.

## Pitch B: the Pathfindr driver

The interesting problem is arithmetic. Pathfindr allows **120 API calls per minute across the whole tenant**. The Edge Agent polls per address on a 10 second interval, which is 6 calls per minute per address, so roughly twenty configured metrics would exhaust a site's entire budget.

So the driver polls collections, not metrics. The asset, runtime and activity endpoints each return the whole estate in one paginated sweep, and every configured metric is served from that cache. API cost becomes a function of estate size and cache window, independent of metric count. Fifty concurrent polls of the same collection make one call.

Also handled: OAuth2 client credentials with refresh, pagination that rebuilds page URLs (the vendor documents that paging links drop query parameters), JSON:API flattening so paths read `$.serialno` rather than `$.data.attributes.serialno`, and truncation that logs loudly rather than presenting a partial estate as a complete one.

Write operations (alerts, launch, recycle) are an explicit no-go. They fire irreversible actions on physical assets and solve a different problem.

## Three decisions worth reviewing

**`srcdoc` was rejected in favour of a shell page.** A srcdoc iframe inherits the embedder's CSP, so hosting driver documents that way would have forced the whole console to allow `'unsafe-inline'` script. Serving the frame from `/driver-ui/host.html` lets nginx give it a separate, tighter policy. This is why there is a shell page at all.

**The console had no CSP whatsoever.** That is pre-existing and independent of this work, but introducing third-party documents is the point at which it stops being acceptable, so one is added here.

**Connection configs are redacted before crossing into the sandbox.** Configs routinely hold credentials: the Pathfindr schema has a `clientSecret` and the stock REST driver has a `password`. Anything the connection schema marks `format: password` arrives as `"__redacted__"`.

## How to test

See the test playbook posted as a comment below. In short:

```sh
cd acs-admin     && npm install && npx vitest run ./test     # 431 pass
cd edge-pathfindr && npm install && npm test                 # 56 pass
cd acs-service-setup && npm install --install-links && node bin/check-yaml.js
```

## Known gaps

- **`npx vite build` fails on `acs-admin`**, and does so on `main` too. `long` is an undeclared transitive dependency of `js-service-client`; the Docker build resolves it via `--install-links`. Not introduced here, but it means the admin bundle was verified by SFC compile check plus unit tests rather than a full local build.
- **Nothing has been run against a real Pathfindr tenant.** The API tests run against a local fake reproducing the documented v5 shapes. Confirming against the live tenant is step 1 of the playbook and should happen before merge.
- **The Manager side has no browser-level test.** The contract logic has 38 unit tests, but the iframe handshake itself is verified manually per the playbook.


AlexGodbehere:
# Review: drivers can ship their own configuration UI, and Pathfindr is the first to do it

**Branch:** `ago/custom-driver-uis`
**Worktree:** `.claude/worktrees/custom-driver-uis`
**PR:** #714
**Base:** `main`

Two things, in that order.

**A driver can now supply its own interface** for the per-metric part of the origin map editor. If a driver definition carries a `ui` key, the Manager renders that document in place of the Address / Path / type / limits fields. If it does not, nothing changes anywhere.

**The Pathfindr driver is the first to use it.** Instead of typing `enviro/SN12345` into a box and `$.latest_temperature` into another, you pick "Temperature" from a list, type the serial off the beacon, and the panel fills in the address, the JSONPath, the Sparkplug type, the unit (°C) and the limits (-20 to 60) together.

### Why it is shaped this way

The scope question that mattered was what a custom UI replaces. Replacing the whole editor would mean every driver author reimplements roughly 3,100 lines of tag tree, schema selection, static values and CSV handling, so the feature would be used exactly once, by us. Replacing only the per-metric region keeps a custom UI to one focused form panel, which is small enough that someone else might actually write one.

The delivery question resolved itself once I looked properly. A third-party driver already has to get its DriverDef into ConfigDB, and `deploy/templates/service-setup.yaml` already mounts an `EXTRA_DUMPS` ConfigMap for exactly that. Shipping a UI is one more key in a file the author was already writing. No Files service upload, no pulling driver images centrally, no new infrastructure.

### Worked example: the arithmetic behind the driver

Pathfindr allows 120 API calls per minute across the whole tenant. The Edge Agent polls per address every 10 seconds, so one address costs 6 calls/min.

| Approach | 20 metrics | 200 metrics |
|---|---|---|
| One call per address | 120 calls/min (at the limit) | 1,200 calls/min (10x over) |
| This driver, 100 assets, 60s cache | 3 calls/min | 3 calls/min |
| This driver, 1000 assets, 60s cache | 30 calls/min | 30 calls/min |

Cost tracks estate size, not metric count. That is the whole reason the driver polls collections and serves metrics from cache.

## Walkthrough

### Custom driver UIs

- Drivers without a `ui` key are completely unaffected. ADS, RTDE, Modbus, BACnet and every custom driver in the field get the fields they get today.
- The document runs in an iframe with `sandbox="allow-scripts"` and deliberately no `allow-same-origin`, so it has an opaque origin and cannot touch the console, its storage or the operator's session.
- It is loaded from `/driver-ui/host.html`, not via `srcdoc`. A srcdoc frame inherits the embedder's CSP, so hosting documents that way would have forced the entire console to permit inline script. The shell page exists solely so nginx can give the frame its own tighter policy.
- That policy is `default-src 'none'`, so a driver UI cannot fetch, XHR, open a WebSocket or load a subresource. The "no live API access" boundary is enforced by the browser rather than by convention.
- Anything the document proposes is filtered through an allowlist and checked against the metric schema before it reaches the model. The Manager stays the only thing that writes to ConfigDB.
- Connection configs are redacted on the way in: anything the schema marks `format: password` arrives as `"__redacted__"`. Pathfindr's `clientSecret` and the REST driver's `password` both qualify.
- Falls back to the standard fields on: no `ui` key, an unrecognised contract version, a document that fails to load, a document that never answers within 5 seconds, or the operator clicking **Use standard fields**. A **Use the driver's interface** control appears to go back.

### The console CSP

- The console had no Content-Security-Policy at all. That is pre-existing, but introducing third-party documents is the point at which it stops being acceptable.
- Two policies now: a strict one for the console, and a separate `default-src 'none'` one for the driver UI frame.
- `connect-src` is left open on the console because service URLs are injected at deploy time by import-meta-env and are not knowable at build time.

### The Pathfindr driver

- Polls collections, not metrics. Fifty metrics backed by the asset collection cost one API call per cache window, not fifty.
- Concurrent polls for the same collection share a single in-flight request.
- Paging links are never followed, because the vendor documents that they drop query parameters. Pages are constructed with our own filters reapplied each time.
- A sweep that hits `maxPages` logs `TRUNCATED` with how much it skipped. A partial estate must never look like a complete one.
- Serial numbers are the only asset identifier exposed, because that is what is printed on the beacon. Internal numeric ids are resolved by the driver.
- Failure mapping is deliberate: a 404 on one address never takes down the connection, while a revoked secret reports as an auth failure rather than going quietly stale.

## How to test

### 0. Setup

```sh
git fetch origin && git checkout ago/custom-driver-uis
```

### 1. Automated tests

```sh
cd acs-admin && npm install && npx vitest run ./test
#  expect: 5 files, 431 tests passing (38 of them new)

cd ../edge-pathfindr && npm install && npm test
#  expect: 56 tests passing

cd ../acs-service-setup && npm install --install-links && node bin/check-yaml.js
#  expect: every dump checked, no errors. This proves the new
#  ACS.Driver.Pathfindr UUID reference resolves.
```

### 2. Confirm the API against the real tenant (do this before merging)

Everything in the driver derives from published documentation. With live credentials:

```sh
curl -s -X POST https://<subdomain>.pathfindr.co.uk/oauth/token \
  -d grant_type=client_credentials \
  -d client_id=<id> -d client_secret=<secret> | jq .

TOKEN=<access_token from above>
curl -s -H "Authorization: Bearer $TOKEN" \
  "https://<subdomain>.pathfindr.co.uk/api/client/v5/assets?page=1" | jq '.meta, .data[0]'
```

Check specifically:
1. `meta.last_page` and `meta.total` are present (pagination depends on them).
2. `data[0].attributes.serialno` exists, and `enviro` / `location_data` / `fluid_level_latest` look as documented.
3. `data[0].attributes.attributes` is the site-defined attribute list.

If any of these differ, the flattening in `lib/api.js` needs adjusting before the driver is useful.

### 3. Regression: existing drivers are untouched

1. Open the Manager, go to an existing Edge Agent with a Modbus, ADS or RTDE connection.
2. Open a device, open the origin map editor, select a metric.
3. **Expect:** exactly what you see today. Address and Path text boxes, no new controls, no "Use standard fields" button.

This is the most important check in the list. A regression here affects every deployment.

### 4. The Pathfindr panel

1. Create a connection with the Pathfindr driver. Fill in base URL, client id and client secret.
2. Create a device on it and open the origin map editor.
3. Select a metric.
4. **Expect:** a "Supplied by the driver" panel with a Data dropdown grouped into Location / Condition / Runtime / Asset / Site, a Serial number field, and a preview showing Address, Path, Type and Unit.
5. Choose **Condition → Temperature** and type a serial.
6. **Expect:** preview shows `enviro/<serial>`, `$.latest_temperature`, `Float`, `°C`.
7. Save, reopen the metric.
8. **Expect:** the panel comes back with Temperature and the serial still selected, not reset to defaults.
9. Choose **Asset → Custom attribute** and type an attribute name with a space in it.
10. **Expect:** path is `$['Your Attribute']`, using bracket notation.

### 5. The escape hatch

1. With the Pathfindr panel showing, click **Use standard fields**.
2. **Expect:** the normal Address / Path / type fields, populated with whatever the panel had set.
3. **Expect:** a **Use the driver's interface** button in the Origin Map header.
4. Click it. **Expect:** the panel returns.

### 6. Containment (the security-relevant checks)

With the Pathfindr panel open, in DevTools:

1. **Console tab:** confirm no CSP violations logged during normal use.
2. **Network tab, filter to Doc:** confirm `/driver-ui/host.html` was fetched, and check its response headers show `Content-Security-Policy: default-src 'none'; script-src 'unsafe-inline'; ...`.
3. Confirm the main document's response headers show the strict policy with `script-src 'self'`.
4. **Elements tab:** confirm the iframe has `sandbox="allow-scripts"` and **not** `allow-same-origin`.
5. In the console, run `document.querySelector('iframe[title*="Driver"]').contentDocument`.
   **Expect:** `null` or a SecurityError. If it returns a document, the sandbox is not doing its job.
6. Confirm the client secret never reaches the frame: in the Network tab there is no request from the frame at all, and if you instrument the panel to log its `init` message, `config.clientSecret` reads `__redacted__`.

### 7. Fallback when a UI misbehaves

Easiest via a temporary ConfigDB edit to the Pathfindr driver definition:

1. Change `ui.version` from `1` to `99`. Reload the editor.
   **Expect:** standard fields, no panel, no error.
2. Restore it, then replace `ui.document` with `<!doctype html><p>silent</p>` (no script, so it never announces itself).
   **Expect:** an empty frame for 5 seconds, then a fall back to standard fields.
3. Restore the original definition.

### 8. The driver against the live tenant

1. Deploy the edge agent with the Pathfindr driver image.
2. Watch the logs: expect `Authenticated to https://<subdomain>...`.
3. Confirm metrics populate in the UNS for a known asset serial.
4. **Rate limiting:** configure 30+ metrics against the same connection, then count requests in the driver logs over a minute. Expect roughly one sweep per `cacheMs`, not one call per metric.
5. **Truncation:** temporarily set `maxPages` to 1 on a tenant with more than 100 assets.
   **Expect:** a `TRUNCATED` log line naming pages fetched vs total. Restore afterwards.
6. **Auth failure:** change the client secret to something wrong.
   **Expect:** the connection reports an auth failure in the Manager, not a silent stall.

## Decisions / open questions

- **`srcdoc` was rejected in favour of a shell page.** This is the least obvious thing in the PR and the reason `public/driver-ui/host.html` exists. A srcdoc frame inherits the parent CSP, so the alternative was `'unsafe-inline'` across the whole console.
- **`npx vite build` fails on `acs-admin`, on `main` as well as here.** `long` is an undeclared transitive dep of `js-service-client`; Docker resolves it via `--install-links`. The admin changes were verified by SFC compile check and unit tests rather than a full local bundle. Worth fixing separately.
- **No browser-level test of the iframe handshake.** The contract has 38 unit tests, but the postMessage exchange itself is only covered manually, by step 6 above.
- **Nothing has touched a real Pathfindr tenant.** Step 2 should happen before merge.
- **The cache window default of 60s is a judgement call.** It suits assets moved by people. A site tracking fast-moving stock will want it lower, and has a schema field to do that.
- **Write operations are a no-go here**, but alerts are the natural follow-up pitch once reads are proven.

## Files of note

**The contract (start here)**
- `acs-admin/src/lib/driver-ui/contract.js` — message shapes, allowlist, validation, redaction
- `acs-admin/test/driver-ui-contract.test.js` — 38 tests, mostly about misbehaving documents

**Manager integration**
- `acs-admin/src/components/EdgeManager/Devices/OriginMapEditor/DriverCustomUI.vue` — the frame host
- `acs-admin/src/components/EdgeManager/Devices/OriginMapEditor/SparkplugMetric.vue` — fallback and toggles
- `acs-admin/public/driver-ui/host.html` — the shell page, and why it exists
- `acs-admin/.docker/nginx.conf` — the two policies

**The driver**
- `edge-pathfindr/lib/api.js` — auth, caching, coalescing, rate limiting, pagination
- `edge-pathfindr/lib/addr.js` — address grammar and the collection-sharing that makes it cheap
- `edge-pathfindr/ui/metric-panel.html` — the panel
- `edge-pathfindr/test/api.test.js` — fake Pathfindr, including the paging quirk

**Registration**
- `acs-service-setup/dumps/edge.yaml` — driver definition with the panel inlined
- `docs/architecture/edge-management/driver-custom-uis.md` — for driver authors

AlexGodbehere:
Interactive version of the test playbook, with checkboxes that persist locally: https://claude.ai/code/artifact/15a6ae3c-0e01-4406-acf5-fc7b6dcbbbc8

AlexGodbehere:
## Verified against the live tenant, and screenshots

Three commits since the original push. The API turned out to disagree with its own documentation in ways that would have shipped silently-empty metrics.

### Screenshots

The real panel, driven through the real shell page and the real contract module under the real nginx policies, captured in a browser. Not mock-ups.

**One choice, five fields set.** Pick "Temperature", type the serial off the beacon, and the address, JSONPath, Sparkplug type, unit and engineering limits are all filled in together:

Temperature

**Battery**, which quietly needs two API calls behind the scenes:

Battery

**A site-defined attribute** whose name contains spaces, so the path needs bracket notation:

Custom attribute

**Validation**, and the known issue below:

Validation

Also in `docs/assets/driver-custom-uis/`: `01-empty.png` and `04-location.png`.

### What the live API exercise found

`tools/catalogue-api.sh` reproduces it; `tools/api-catalogue.txt` is the output.

| Finding | Consequence |
|---|---|
| The asset collection omits the site-defined attribute list unless `include=attributes` is requested. `enviro` is **not** a valid include and is rejected. | `attrs/<serial>` could never have returned anything. Fixed. |
| The enviro history endpoint ignores `per_page` and forces 1440 rows a page. One tag had 11,089 records over 8 pages. | A single temperature metric would have spent 8 API calls and downloaded 11k rows **per cache window**. History addresses now fetch page one and stop. |
| `enviro` and `fluid_level_latest` never appear on the asset record. | Temperature and humidity pointed at fields that do not exist. Both now read the newest enviro history record. |
| `beacon_info` (battery, last heartbeat) exists on the single-asset endpoint and nowhere else. Undocumented. | Battery is the most useful field on a BLE tag, since a flat beacon stops reporting everything. Added as `beacon/<serial>`. |
| Bad credentials return `500 {"message":""}`, not `401 invalid_client`. | A mistyped secret would have reported as a connection failure and sent someone hunting firewalls. Fixed. |
| `expires_in` is **1200** (20 min), not the documented 63115200 (2 years). | The refresh path runs several times an hour rather than never, so it now has direct test coverage. |
| The runtime endpoints answer `422 invalid_gps_tracker` for any non-GPS asset. | Routine for BLE tags, already non-fatal. The panel now says so instead of offering entries that cannot work. |

### Two bugs the screenshot harness found

Both would have shown an operator an empty panel until the 5-second fallback fired:

- **The shell announced itself once**, the instant it parsed. Any embedder attaching its listener a tick later (a deferred or module script) missed it entirely and the exchange never started. It now repeats until the document arrives.
- **A failed `document.write` reported nothing**, leaving the embedder to time out. It now reports, and the Manager falls back immediately.

Plus a cosmetic one: the frame grew without bound, because `report_height` measured `documentElement.scrollHeight`, which returns `max(viewport, content)` — so it measured back whatever height it had just been given.

### Known issue, deliberately not fixed here

Visible in the validation screenshot: the panel stops proposing when input goes invalid, but never **retracts** its last good proposal. Type a valid serial, then break it, and the metric keeps the old address while the UI says it is invalid.

The obvious fix, proposing nulls to clear the fields, would delete the metric mid-edit every time someone clears the field to retype, because the editor removes metrics with no Address, Path or Value. That needs a considered fix rather than a rushed one.

### Test counts

| Suite | Tests |
|---|---|
| `acs-admin` | 431 pass |
| `edge-pathfindr` | 71 pass (was 41) |
| `acs-service-setup` dumps lint | clean |

The mechanism is now verified end-to-end in a real browser: the panel loads inside the sandboxed frame, through the shell, under `default-src 'none'`, and proposes validated values back.

PR #727, opened 2026-08-20
java-service-client: stop caching failed Directory lookups forever

AlexGodbehere:
## Summary

Fixes a failure mode in the Java service client's Directory discovery that can take down all MQTT data flow until the broker is restarted.

If a client using `java-service-client` (notably the HiveMQ Kerberos auth extension) starts while the Directory cannot reach its database, the failed lookup is returned as an empty URL set and cached forever. The broker then rejects every MQTT CONNECT with "Bad User Name or Password" because it can never find the Auth service to fetch ACLs, even after the Directory recovers. This was observed in production after a node reschedule moved the broker and the Directory's Postgres at the same time.

Three changes in `lib/java-service-client`:

1. **`FPDirectory.getServiceURLs`**: a non-OK Directory response is now an error, not an empty array. Errors are never cached, so the lookup is retried.
2. **`RequestCache`**: supports an optional expiry and a cacheability predicate. Expired entries are refetched on the next request; if the refetch fails the stale value is served, so a Directory outage does not invalidate URLs a service already holds. Entries set with `put()` (explicit config such as `directory_url`) never expire. Also fixes a latent race where a synchronously-completing source leaked a completed promise into the in-flight map.
3. **`FPDiscovery`**: caches lookups for five minutes and never caches an empty result. The long-standing `XXX` comment about indefinite caching being incorrect is resolved.

## How to test

1. `cd lib/java-service-client && mvn test` - 8 tests pass, including new `RequestCacheTest` coverage for error handling, empty results, expiry, stale fallback, and pinned entries.
2. `cd acs-mqtt && mvn compile` against the installed library - compiles unchanged.
3. Manual reproduction: start the MQTT broker while the Directory returns 500 (for example while its database is unreachable). Previously every MQTT client is rejected forever; with this change the broker recovers as soon as the Directory does, on the next connection attempt.

## Release notes

- Fixed: an MQTT broker (or any Java service client) that started while the Directory was unavailable would cache the failed service lookup forever and reject all connections until restarted. Service URL lookups now expire after five minutes, failed and empty lookups are retried, and a known-good URL is kept if the Directory is temporarily unreachable.

### DIFF
--- a/lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/FPDirectory.java
+++ b/lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/FPDirectory.java
@@ -51,14 +51,16 @@ public class FPDirectory {
                 .appendPath(service.toString())
             )
             .fetch()
-            .map(res -> res.ifOk()
-                .flatMap(r -> r.getBody())
+            .flatMap(res -> res.ifOk()
+                .flatMap(r -> r.getBodyArray())
+                .map(Single::just)
                 .orElseGet(() -> {
                     log.error("Can't find {} via the Directory: {}",
                         service, res.getCode());
-                    return new JSONArray();
+                    return Single.error(new Exception(
+                        "Directory lookup of %s failed: %s"
+                            .formatted(service, res.getCode())));
                 }))
-            .cast(JSONArray.class)
             .flatMapObservable(Observable::fromIterable)
             .doOnNext(o -> log.info("Service URL: {}", o))
             .cast(JSONObject.class)
--- a/lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/FPDiscovery.java
+++ b/lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/FPDiscovery.java
@@ -6,6 +6,7 @@
 package uk.co.amrc.factoryplus.client;
 
 import java.net.URI;
+import java.time.Duration;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -21,12 +22,14 @@ import io.reactivex.rxjava3.core.*;
 
 import uk.co.amrc.factoryplus.http.*;
 
-/* XXX The JS client performs a (cached) fetch every time, with a Map of
- * overrides. This client caches the responses from the Directory
- * indefinitely, which is definitely incorrect. However, we perform
- * not-entirely-trivial processing on the response, so even if there
- * will be no network activity I would like to know that I got a 304 and
- * can reuse the existing derived value... */
+/* Responses from the Directory are cached for a limited time, after
+ * which they are refetched on the next request. If the refetch fails
+ * the stale value is used, so a Directory outage does not break
+ * services which already know where to go. Empty responses and errors
+ * are never cached; previously a failed lookup was cached as an empty
+ * set indefinitely, which left long-running services (notably the MQTT
+ * broker auth plugin) permanently broken if they started up while the
+ * Directory was unavailable. */
 
 /** Service discovery.
  *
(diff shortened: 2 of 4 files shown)


AlexGodbehere:
# Review: stop the Java client caching failed Directory lookups forever

**Branch:** `fix-discovery-cache`
**Worktree:** `.claude/worktrees/fix-discovery-cache`
**PR:** #727

A broker (or any Java service client) that started while the Directory was unavailable would cache the failed Auth service lookup as an empty set, forever. Every MQTT CONNECT then failed with "Bad User Name or Password" until someone restarted the broker, because the auth extension could never find the Auth service to fetch ACLs. The Directory recovering did not help; the client never asked again.

Three behavioural changes:

1. A non-OK Directory response is now an error, not an empty result. Errors were already never cached, so the next connection retries the lookup.
2. Cached service URLs expire after 5 minutes. On the next request after expiry the client refetches; if that refetch fails, it keeps using the stale URL. So a Directory outage no longer breaks a service that already knows where to go, in either direction.
3. Empty lookup results are never cached, so a service with no registrations yet is re-checked on every request.

Worked example (the incident): broker starts at T+0 while the Directory's database is unreachable, lookup returns 500. Before: 500 becomes an empty set, cached forever; all clients rejected until a manual broker restart. After: 500 is an error, not cached; the first CONNECT after the Directory recovers triggers a fresh lookup and authentication succeeds. Steady state: the URL is refetched at most every 5 minutes, and if the Directory is down at that moment the old URL is kept.

## Walkthrough

### Directory lookup errors (`FPDirectory.java`)
- `getServiceURLs` propagates a non-OK response as `Single.error` with the status code in the message, instead of logging and returning `[]`.
- Uses the typed `getBodyArray()` instead of `getBody()` plus a cast.

### Cache semantics (`RequestCache.java`)
- New constructor takes an expiry `Duration` and a `Predicate<Value>` for cacheability; the old constructor keeps the old behaviour (cache everything, forever) and is still used for bearer tokens.
- Entries record an expiry instant; `put()` entries (explicit config like `directory_url`) get a null expiry and never expire.
- When a stale entry exists and the refetch errors, the stale value is returned (`onErrorReturnItem`).
- Fixed a latent race inherited from the old code: the fetch was subscribed inside `computeIfAbsent`, so a source that completed synchronously removed itself from the in-flight map before insertion and leaked a completed promise that was handed to all future callers. The fetch is now deferred to the first subscriber via `cache()`, which is always after insertion. The new unit tests caught this.

### Discovery policy (`FPDiscovery.java`)
- 5-minute expiry, never cache empty sets.
- Replaced the `XXX` comment admitting the indefinite cache was "definitely incorrect" with a description of the new behaviour.

## How to test

1. `cd lib/java-service-client && mvn test` - expect 8 passing tests (7 new in `RequestCacheTest`).
2. `cd lib/java-service-client && mvn install -DskipTests && cd ../../acs-mqtt && mvn compile` - the HiveMQ extension compiles unchanged.
3. Integration: start the MQTT broker while the Directory returns 500 (easiest: scale the Directory's Postgres down). Attempt an MQTT connection, see it rejected. Restore the Directory. Reconnect: the connection now succeeds without restarting the broker.

## Decisions / open questions

- 5 minutes is a judgement call. It bounds how long a moved service URL stays stale without hammering the Directory from every connection attempt.
- Stale-on-refetch-failure means a genuinely deregistered service keeps resolving to its old URL while the Directory is down. That trade was taken deliberately: a wrong URL fails loudly at connect time, whereas the alternative reintroduces the incident.
- Bearer-token caching in `FPHttpClient` intentionally keeps the old no-expiry constructor; tokens are invalidated on 401 via `BadToken` already.
- `remove()` now compares the caller's value against the entry's wrapped value, preserving the old remove-by-value contract used by token invalidation.
- The JS service client has different (per-fetch) discovery semantics and is not touched.

## Files of note

- `lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/FPDirectory.java` - error on failed lookups
- `lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/RequestCache.java` - expiry, cacheability, stale fallback, in-flight race fix
- `lib/java-service-client/src/main/java/uk/co/amrc/factoryplus/client/FPDiscovery.java` - discovery cache policy
- `lib/java-service-client/src/test/java/uk/co/amrc/factoryplus/client/RequestCacheTest.java` - new tests

PR #733, opened 2026-08-21
Za/files - fixed file write abortion by request timeout

amrc-za:
Fixed node's default 5 minute request timeout aborting file write for large files:
- lib/js-service-api/lib/webapi.js — added an opt-in request_timeout option.
- acs-files/bin/api.js — passes request_timeout
- deploy/templates/files/files.yaml — added a REQUEST_TIMEOUT env var
- deploy/values.yaml — added files.requestTimeout: 0 (0 disables Node's request timeout entirely, so an upload can take as long as it needs)


djnewbould (commented):
The changes to the service api all look good. I can't merge it before the open metadata stuff is merged. It might be worth basing the last three commits off of main so we can merge it sooner.

PR #734, opened 2026-08-21
Za/openmetadata 

amrc-za:
Adds OpenMetadata as a new optional ACS service, deployed via the upstream openmetadata and openmetadata-dependencies Helm charts (v1.13.3).

The upstream images for OpenMetadata's server, its ingestion/Airflow component, and OpenSearch (used as OpenMetadata's search backend) all bundle outdated log4j-core/log4j-api jars with known CVEs. Rather than wait on upstream, we build three patched images that swap in log4j 2.26.1 and use them in place of the stock images.


deploy/templates/openmetadata/openmetadata-logs.pvc.yaml
djnewbould, line 1: Should this be wrapped with {{ if .Values.openmetadata.enabled }}? 
I think they're missing `namespace: {{ .Release.Namespace }} `too

deploy/templates/openmetadata/openmetadata-logs.pvc.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: acs-dags
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
+---
```
djnewbould, line 24: Is this one redundant? I can't see where its used anywhere

acs-service-setup/dumps/openmetadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+      subclassOf:
+        - !u Auth.Class.Permission
+
+# ---
+# service: !u UUIDs.Service.Authentication
+# version: 2
+# grants:
+#   !u Openmetadata.Requirement.ServiceRole:
+#     !u UUIDs.Permission.Directory.AdvertiseService:
+#       !u Openmetadata.Service.Openmetadata: false
```
djnewbould, line 27: Does it need this grant or is it fine without it?

djnewbould (changes requested):
It all looks good! I've just left a couple of small comments.

deploy/templates/openmetadata/openmetadata-logs.pvc.yaml
amrc-za: this is used by api-server dependency, though not explicitly mentioned.

deploy/templates/openmetadata/openmetadata-logs.pvc.yaml
amrc-za: wrapped and added namespace

acs-service-setup/dumps/openmetadata.yaml
amrc-za: remove, at least for now

PR #742, opened 2026-09-02
files - fixed large files aborting due to request timeout

amrc-za:
Fixed node's default 5 minute request timeout aborting file write for large files:

lib/js-service-api/lib/webapi.js — added an opt-in request_timeout option.
acs-files/bin/api.js — passes request_timeout
deploy/templates/files/files.yaml — added a REQUEST_TIMEOUT env var
deploy/values.yaml — added files.requestTimeout: 0 (0 disables Node's request timeout entirely, so an upload can take as long as it needs)

### DIFF
--- a/deploy/values.yaml
+++ b/deploy/values.yaml
@@ -127,7 +127,8 @@ files:
   enabled: true
   image:
     repository: acs-files
-  bodyLimit: 500kb
+  bodyLimit: 3gb
+  requestTimeout: 0 # 0 disables HTTP request timeout, allowing uploads of any size/duration
   storage: 20Gi
 
 monitor:
--- a/deploy/values.yaml
+++ b/deploy/values.yaml
@@ -128,7 +128,6 @@ files:
   image:
     repository: acs-files
   bodyLimit: 3gb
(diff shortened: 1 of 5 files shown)


deploy/values.yaml
```diff
@@ -127,7 +127,8 @@ files:
   enabled: true
   image:
     repository: acs-files
-  bodyLimit: 500kb
+  bodyLimit: 3gb
+  requestTimeout: 0 # 0 disables HTTP request timeout, allowing uploads of any size/duration
```
djnewbould: We might want to enable a request timeout as this is the default value. Maybe 120s or nodes default 5 minute timeout? Users can disable it in their own values file if they want to support larger file uploads.

djnewbould (changes requested):

deploy/values.yaml
amrc-za: Removed requestTimeout from default values completely

amrc-za: this makes the timeout to be Node's default.

djnewbould (approved):

PR #744, opened 2026-09-07
Za/central tdms summariser

amrc-za:
New acs-file-summariser service that watches for specific file types (currently only implemented for TDMS) via ConfigDB NotifyAPI. Upon receiving a notification about new file upload the service creates a summary of the TDMS file and saves it to InfluxDB.


deploy/values.yaml
```diff
@@ -130,6 +130,32 @@ files:
   bodyLimit: 3gb
   storage: 20Gi
 
+fileSummariser:
+  enabled: true
```
djnewbould: I think we'd want this value to be set to false by default. I don't think every ACS deployment would need to have this running, I think it'd be an edge case

acs-file-summariser/Dockerfile
```diff
@@ -0,0 +1,46 @@
+# syntax=docker/dockerfile:1
+
+ARG base_version
+ARG base_prefix=ghcr.io/amrc-factoryplus/acs-base
+
+FROM ${base_prefix}-js-build:${base_version} AS build
+ARG revision=unknown
```
djnewbould, line 7: Should the revision number be set to something or is it ok being unknown?

acs-file-summariser/lib/influx.js
```diff
@@ -0,0 +1,65 @@
...
+         * A single file's summary can be a lot of points, so the write
+         * API is left to auto-flush by size (`batch_size`) and by time
+         * (`flush_interval`, a safety net for a slow-trickling file) as
+         * well as being flushed explicitly once a file completes - see
+         * Job.run(). */
+        this.write_api = client.getWriteApi(opts.org, opts.bucket, "ns", {
+            batchSize:      opts.batch_size,
+            flushInterval:  opts.flush_interval,
+            maxBufferLines: 30_000,
+            maxRetries:     0,
```
djnewbould, line 37: If an auto-flush fails, data could be lost without any indication. We could just log any failures with:
```
this.write_api = client.getWriteApi(opts.org, opts.bucket, "ns", {
    batchSize:      opts.batch_size,
    flushInterval:  opts.flush_interval,
    maxBufferLines: 30_000,
    maxRetries:     0,
    writeFailed:    (error, lines, attempt, expires) =>
        this.log("Failed to write %d point(s) to InfluxDB: %s", lines.length, error),
});
```

djnewbould (changes requested):
Looks good and well documented! I've just left a couple of comments. I think it also needs adding to the github build action

deploy/values.yaml
amrc-za: Disabled by default

acs-file-summariser/Dockerfile
amrc-za: added git version to both fileSummariser and files services

amrc-za:
I have also added optional `device_uuid ` to be recorded as part of Files metadata in ConfigDB.
File Summariser reads the `device_uuid `and stores it as `topLevelInstance `tag in influxdb.

djnewbould (approved):

PR #747, opened 2026-09-15
Za/openmetadata

amrc-za:
Openmetadata 2.0.1 is added as a new ACS service through official openmetadata helm charts.


deploy/values.yaml
```diff
@@ -715,3 +715,161 @@ influxdb2:
...
+        airflow:
+          auth:
+            enabled: true
+            username: admin
+            password:
+              secretRef: airflow-secrets
+              secretKey: openmetadata-airflow-password
+
+openmetadata-dependencies:
+  enabled: true
```
djnewbould: I think openmetadata and openmetadata-dependencies should be false by default, the stack it deploys is pretty resource heavy and I don't think every deployment would need it.

.gitignore
```diff
@@ -4,7 +4,9 @@ config.mk
 .env
 
 */.npm_install_done
-.vscode
+.vscode/
+.claude/
+.devcontainer/
```
djnewbould: Do these two folders need to be git ignored? we ignore specific files in these folders already

djnewbould (changes requested):
looks good, just requested a couple of small changes

.gitignore
amrc-za: removed from gitignore

deploy/values.yaml
amrc-za: disabled by default

djnewbould (approved):

